PR #24, opened 2026-07-23
fix(browser): clear stale Chromium Singleton locks before relaunch

### LINKED ISSUE #22: Playwright browser context lock persists on process SIGINT / tool failure, causing subsequent browser runs to fail
Environment
- OS: Linux x64 (Ubuntu 24.04 LTS) / macOS Sonoma (Apple Silicon)
- Runtime / Binary: Standalone SEA / Node.js v20.x
- atomic-agent version:`v0.4.x` (main branch)
- Backend Model: Local llama.cpp (q4_k_m with TurboQuant KV-cache)

-Summary
When atomic-agent is executing a multi-step browser automation task via playwright-core and the process is unexpectedly interrupted (e.g., user sends Ctrl+C / SIGINT, or a tool invocation times out), the browser profile lock file under <stateDir>/browser/ is not cleaned up gracefully. 

On subsequent runs, any tool call attempting to invoke browser.navigate or browser.inspect fails immediately with an unhandled rejection, as the agent assumes another browser instance is currently active.

Steps to Reproduce
1. Start an atomic-agent task requiring web browsing:
   ```bash
   atomic-agent run "Navigate to news.ycombinator.com and extract top 3 posts"

Bartok9:
## Summary
Hard kills / SIGINT before `PlaywrightBackend.shutdown()` finishes leave Chromium `SingletonLock` (and friends) under the persistent `user-data-dir`. The next `browser.*` launch then fails until manual cleanup.

## Fix
- New `clearStaleChromeLocks(userDataDir)` with PID-aware policy:
  - dead owner PID → remove locks
  - live Chromium-family holder → keep
  - orphan companion locks without SingletonLock → remove
- Called in `launchOwnChrome()` before `decorateChromeProfile` / spawn
- Unit tests for parse + clear paths

## Test plan
- [x] `vitest run src/tools/browser/clear-stale-chrome-locks.test.ts`
- [ ] Manual: start browser task, Ctrl+C mid-run, relaunch browser tool without deleting profile

Fixes #22

Agent-Owner: sera

### DIFF
--- a/src/tools/browser/clear-stale-chrome-locks.test.ts
+++ b/src/tools/browser/clear-stale-chrome-locks.test.ts
@@ -1,0 +1,96 @@
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import {
+  CHROME_LOCK_BASENAMES,
+  clearStaleChromeLocks,
+  readChromeSingletonLockPid,
+} from "./clear-stale-chrome-locks.js";
+
+describe("readChromeSingletonLockPid", () => {
+  let dir: string;
+  beforeEach(() => {
+    dir = fs.mkdtempSync(path.join(os.tmpdir(), "atomic-locks-"));
+  });
+  afterEach(() => {
+    fs.rmSync(dir, { recursive: true, force: true });
+  });
+
+  it("parses symlink targets of the form host-pid", () => {
+    fs.symlinkSync("127.0.0.1-4242", path.join(dir, "SingletonLock"));
+    expect(readChromeSingletonLockPid(dir)).toBe(4242);
+  });
+
+  it("parses plain text pid files", () => {
+    fs.writeFileSync(path.join(dir, "SingletonLock"), "99999\n");
+    expect(readChromeSingletonLockPid(dir)).toBe(99999);
+  });
+
+  it("returns null when missing", () => {
+    expect(readChromeSingletonLockPid(dir)).toBeNull();
+  });
+});
+
+describe("clearStaleChromeLocks", () => {
+  let dir: string;
+  beforeEach(() => {
+    dir = fs.mkdtempSync(path.join(os.tmpdir(), "atomic-locks-"));
+  });
+  afterEach(() => {
+    fs.rmSync(dir, { recursive: true, force: true });
+  });
+
+  function plantLocks(pidPayload: string | null): void {
+    for (const base of CHROME_LOCK_BASENAMES) {
+      if (base === "SingletonLock" && pidPayload !== null) {
+        fs.symlinkSync(pidPayload, path.join(dir, base));
+      } else if (base !== "SingletonLock") {
+        fs.writeFileSync(path.join(dir, base), "x");
+      }
+    }
+  }
+
+  it("removes companion locks when SingletonLock is absent", () => {
+    fs.writeFileSync(path.join(dir, "SingletonSocket"), "x");
+    fs.writeFileSync(path.join(dir, "DevToolsActivePort"), "9222");
+    const result = clearStaleChromeLocks(dir);
+    expect(result.removed).toEqual(
+      expect.arrayContaining(["SingletonSocket", "DevToolsActivePort"]),
+    );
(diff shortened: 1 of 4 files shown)


sosidudku1:
Thanks @Bartok9. The problem is real: the profile dir is persistent, nothing in the tree touches `SingletonLock` today, and `run-agent.ts:294` registers SIGINT with `process.once`, so a second Ctrl+C or a hard kill skips shutdown entirely. PID-aware cleanup is the right shape for this.

Two things I would like changed before this lands.

**The unknown-owner path clears instead of keeping.** `processLooksLikeChromium` returns `false` from its catch, and the caller treats `false` as "not chromium, safe to clear". So when the probe itself fails (no `/proc` access, `ps` timeout, unexpected platform) a live browser's locks get removed, which is the opposite of the comment three lines above it saying unknown should keep. Distinguishing "probed and it is not Chromium" from "could not probe" would fix it.

**The dead-pid test can pass without asserting anything.** It returns early if the PID turns out to be alive, so on a machine where that happens the test is green while proving nothing. Injecting the liveness check would make it deterministic.

Separately, and not your fault: the `index.ts` hunk here re-exports `DEFAULT_ARIA_MAX_CHARS` and `packLinesToCharBudget`, which do not exist on main and are only introduced by #25. As it stands this branch does not build on its own. Either drop those two lines or we land #25 first, whichever you prefer.

Your test plan also leaves the manual case unchecked. That is the scenario I care most about here, so if you get a chance to run it, say what you saw.

Ooooze:
Thanks for tackling this — the PID-aware lock policy is a good approach. Two things need fixing before this can merge:

1. **Build is broken on this branch.** `tsc` fails: `index.ts` re-exports `DEFAULT_ARIA_MAX_CHARS` and `packLinesToCharBudget` from `aria-compressor.ts`, but this branch's `aria-compressor.ts` doesn't define them (that code lives in #25). Looks like a rebase picked up a hunk without its dependency — could you rebase cleanly on `main` and drop whatever pulled that in?
2. **Fail-path inverts the stated policy.** In `processLooksLikeChromium`, when `ps`/`/proc` lookup throws (permission denied, sandboxed environment, ENOENT), the function returns `false` — which `clearStaleChromeLocks` then reads as "not Chromium → remove the lock". The comment in the code says the opposite should happen ("on failure → unknown → keep locks"). As written, a live browser whose process check fails for an unrelated reason could have its locks pulled out from under it. Could you flip that to fail-open (keep locks on lookup failure) and add a test for it?

Happy to take another look once those are sorted.

Bartok9:
Thanks @sosidudku1 @Ooooze — all three addressed in 48c2453.

**Build.** Rebased cleanly on main and dropped the `index.ts` re-exports of `DEFAULT_ARIA_MAX_CHARS` / `packLinesToCharBudget` (those belong to #25). This branch now builds on its own — `tsc --noEmit` clean.

**Fail-path inversion.** `processLooksLikeChromium` is now tri-state — `"chromium" | "other" | "unknown"`. A successful probe that finds a non-browser returns `"other"` (clear); a probe that *can't run* (no `/proc`, `ps` timeout, unexpected platform, Windows) returns `"unknown"`, which the caller treats as keep-locks. So a lookup failure can no longer be read as "not chromium → remove" and pull locks from a live browser.

**Deterministic test.** `clearStaleChromeLocks` now takes `isAlive`/`probe` overrides. The dead-pid test injects `isAlive: () => false` so it asserts removal every run instead of early-returning. Added three more: live-chromium (keep), live-other (clear), and fail-open unknown (keep). 10/10 green.

**Manual case.** Ran it: SingletonLock symlink → dead PID + SingletonSocket/Cookie/DevToolsActivePort companions. Parsed PID correctly, then removed all four with reason `pid … not alive`; SingletonLock gone afterward. That's the SIGINT-skipped / kill -9 relaunch path this fixes.

Agreed on landing #25 first.

Ooooze:
Confirmed — builds clean now, the tri-state processLooksLikeChromium (chromium/other/unknown) correctly fails open on a failed probe, and the dead-pid test is deterministic via the isAlive injection. 10/10 green. Merging, thanks for addressing both rounds of feedback in one pass!

Issue #21, opened 2026-07-20
Unpruned HTML trees in browser.inspect tool cause context limit exhaustion on large DOMs


CALEOT-tech:
Environment
-OS:Cross-platform (Linux / macOS / Windows)
-Runtime:Node.js v20.x / Standalone Binary
-atomic-agent version:** `v0.4.x` (main branch)
-Model Provider / Backend:** Ollama / llama.cpp / Anthropic API

Summary
When the agent invokes the `browser.inspect` tool on web pages with dense or dynamic DOM structures (e.g., complex dashboards, Webpack single-page apps, or heavily nested modern frameworks), the tool returns raw or insufficiently sanitized HTML snapshot trees.

This injects tens of thousands of redundant tokens (CSS styles, SVG paths, data attributes, hidden script tags) directly into the agent's context window. On local models with smaller context limits (e.g., 8k–16k tokens), this instantly causes context window overflow and causes the agent loop to crash or lose earlier prompt instructions.

Steps to Reproduce
1. Run `atomic-agent` and instruct it to inspect or interact with a heavy web page:
   ```bash
   atomic-agent run "Go to [github.com/trending](https://github.com/trending) and give me the top 5 repository descriptions."

Ooooze:
Closing this out. #25 merged and covers the remaining case — the crude version (excessive line count) was already handled on main, and #25 adds a character-budget pack (default 24k chars) so dense short-line SPA trees that pass the line cap but are still individually huge no longer blow past small-model context windows. Thanks for the report!

PR #25, opened 2026-07-23
fix(browser): bound ARIA snapshots by character budget

Bartok9:
## Summary
`summariseAriaSnapshot` only capped **lines**. Dense short-line SPA trees still push tens of kB into small local-model contexts (8k–16k), which is the failure mode described in #21.

## Fix
- Order-preserving `maxChars` pack (default **24_000**) applied after noise drop, before the line cap
- Size-truncation footer so the agent knows to scroll/navigate
- `maxChars: 0` / non-finite disables the budget
- Unit coverage for the pack footer path

## Test plan
- [x] `vitest run src/tools/browser/browser-tools.test.ts`
- [ ] Optional: `browser.read_aria` on a heavy SPA and confirm footer + smaller prompt payload

Addresses #21

Agent-Owner: sera


sosidudku1:
Thanks @Bartok9. The line cap alone cannot see how wide a line is, so a char budget is the right second axis. Worth noting for anyone reading #21 later: the crude version of that problem is already fixed on main, and what your patch covers is the remaining case, short dense lines that pass the line cap.

This is queued for merge review. Sorry for the slow first response, we are going through the whole browser batch this week.

Two things I noticed while reading it.

`maxChars: 0.4` passes the `<= 0` guard and then `Math.trunc` makes it `0`, which returns an empty array and wipes the snapshot. `Math.max(1, Math.trunc(...))` closes it.

The new test passes `dropNoise: false`, but `playwright-backend.ts:126` is the only real call site and passes no options, so noise dropping and the char budget always run together there. A test with both defaults would cover what actually executes.

Also, minor: the comment above `packLinesToCharBudget` describes scoring lines with refs higher, but it packs in order and cuts the tail. The behaviour is fine, the comment just promises more than it does.

Merge order: #24 re-exports two symbols from here without introducing them, so this one goes first.

Bartok9:
Thanks @sosidudku1, all three addressed in 9d2aa43:

- **maxChars 0.4 wipe**: now `Math.max(1, Math.trunc(...))`, so a sub-1 budget truncates to one char instead of returning an empty array.
- **defaults coverage**: added a test that exercises the char budget and noise dropping together (the playwright-backend path where both run with no options), plus a fractional-maxChars regression test.
- **comment**: corrected — `packLinesToCharBudget` packs positionally and cuts the tail; no ref/name scoring. The comment now says exactly that.

31/31 green. Agreed on merge order behind #24.

Ooooze:
The character-budget approach is a solid fix for the context-exhaustion problem in #21 — order-preserving pack after noise-drop, before the line cap, with a clear truncation footer. Two cleanup items before merge:

1. **`pnpm-lock.yaml` shouldn't be in this diff.** This repo uses npm (`package-lock.json`), not pnpm — the new 5097-line `pnpm-lock.yaml` looks like it got committed from a local `pnpm install` and isn't related to the ARIA fix. Could you drop it from the branch?
2. **Test location.** The new tests landed in `browser-tools.test.ts`, which is already over the repo's 300-line-per-file guideline even before this PR's additions. Would you mind splitting the new ARIA-budget cases into a dedicated `aria-compressor.test.ts`?

Logic itself looks correct and `tsc`/tests are clean otherwise — thanks for the fix!

Bartok9:
Thanks @Ooooze — both addressed:

1. **pnpm-lock.yaml dropped.** It got committed from a local `pnpm install`; the repo uses npm/`package-lock.json`. Rebased cleanly on main and removed it — no longer in the diff.
2. **Test location.** Moved the whole `summariseAriaSnapshot` suite out of `browser-tools.test.ts` (617 → 487 lines) into a dedicated `aria-compressor.test.ts`. The ARIA-budget cases live there now.

36/36 green, `tsc` clean. Diff is now just `aria-compressor.ts`, the new test file, and the `index.ts` re-export.

Ooooze:
Confirmed — pnpm-lock.yaml is gone and the ARIA-budget suite now lives in its own file. 36/36 green, tsc clean. Merging, thanks for the quick turnaround!

PR #26, opened 2026-07-23
fix(local-llm): detect Apple Metal GPUs in managed device auto-pick

Bartok9:
## Summary
On Apple Silicon, managed mode often printed **`auto → CPU (no GPU detected)`** even when Metal was available, because `parseListDevices` only accepted canonical `VulkanN:` / `CUDAN:` rows.

## Fix
- Parse the real Metal device-table row verbatim (`MTL0: Apple M1 Max (…)`) — id is **read from the table**, never constructed, so no fabricated id can reach `--device` and fail the daemon health poll
- Treat Apple Silicon descriptions as preferred discrete-class devices in `pickBestDevice`
- Merge duplicate rows (later memory-bearing row wins) so a figure-less init line seen first can't drop auto context sizing to the 16384 fallback

## Test plan
- [x] `vitest run src/local-llm/gpu-devices.test.ts` (17)
- [x] On macOS (M1 Max): `llama-server --list-devices` → `MTL0`; `--device MTL0` addressable, `--device Metal0` rejected — parser returns the addressable `MTL0`

Agent-Owner: sera

### DIFF
--- a/src/local-llm/gpu-devices.test.ts
+++ b/src/local-llm/gpu-devices.test.ts
@@ -60,6 +60,42 @@ describe("parseListDevices", () => {
     expect(parseListDevices("")).toEqual([]);
     expect(parseListDevices("no devices here")).toEqual([]);
   });
+
+  it("parses canonical MetalN device lines (Apple Silicon)", () => {
+    const out = [
+      "Available devices:",
+      "  Metal0: Apple M3 Ultra (98304 MiB, 90000 MiB free)",
+    ].join("\n");
+    expect(parseListDevices(out)).toEqual<GpuDevice[]>([
+      {
+        id: "Metal0",
+        description: "Apple M3 Ultra",
+        totalMemMiB: 98304,
+        freeMemMiB: 90000,
+      },
+    ]);
+  });
+
+  it("normalises 'GPU 0:' Metal table rows to Metal0", () => {
+    const out = "GPU 0: Apple M2 Max (id: 0, buffer size: 64 GiB)";
+    const devices = parseListDevices(out);
+    expect(devices).toHaveLength(1);
+    expect(devices[0]?.id).toBe("Metal0");
+    expect(devices[0]?.description).toMatch(/Apple M2 Max/i);
+  });
+
+  it("parses ggml-metal device log lines", () => {
+    const out = "ggml-metal : device 0 = Apple M3 Ultra";
+    const devices = parseListDevices(out);
+    expect(devices).toEqual<GpuDevice[]>([
+      {
+        id: "Metal0",
+        description: "Apple M3 Ultra",
+        totalMemMiB: 0,
+        freeMemMiB: 0,
+      },
+    ]);
+  });
 });
 
 describe("pickBestDevice", () => {
@@ -170,6 +206,20 @@ describe("pickBestDevice", () => {
     ];
     expect(pickBestDevice(devices)).toBe("Vulkan0");
   });
+
+
+  it("treats Apple Silicon Metal devices as auto-pickable (not discarded)", () => {
+    const devices: GpuDevice[] = [
+      {
+        id: "Metal0",
+        description: "Apple M3 Ultra",
+        totalMemMiB: 98304,
+        freeMemMiB: 90000,
+      },
+    ];
+    expect(pickBestDevice(devices)).toBe("Metal0");
+  });
+
 });
 
 describe("resolveManagedDevice", () => {
--- a/src/local-llm/gpu-devices.test.ts
+++ b/src/local-llm/gpu-devices.test.ts
@@ -61,40 +61,42 @@ describe("parseListDevices", () => {
     expect(parseListDevices("no devices here")).toEqual([]);
   });
 
-  it("parses canonical MetalN device lines (Apple Silicon)", () => {
+  it("parses the real Apple Silicon Metal row verbatim (MTL0, from an M1 Max)", () => {
+    // Verbatim from `llama-server --list-devices` on an Apple M1 Max.
+    // The addressable --device id is MTL0, not Metal0 — read it out of
+    // the table rather than constructing it.
     const out = [
       "Available devices:",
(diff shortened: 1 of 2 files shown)


sosidudku1:
Thanks @Bartok9. Third one I have looked at today, and this is the one I would not have found on my own, since none of us runs the managed path on Apple Silicon day to day.

Before this moves, I would like to understand one thing. Two of the three parse branches construct the device id rather than reading it out of the output, and that id goes straight into `--device` at spawn time with no validation in between. If `Metal0` is not what the backend actually answers to, the daemon fails its health poll and the operator gets a 30 second timeout and a log tail instead of a clear error. Your macOS line is still unticked, and one run of `models devices` on real hardware would tell us whether the constructed ids are addressable. If you have already run it and it worked, just say so and that is enough for me.

Second, smaller: the `seen` dedup keeps whichever line arrives first rather than whichever carries more data. On builds that print the canonical table to stdout this never matters. But your own comment notes some builds put the table on stderr, and there an init line seen first would win and the row with the real memory figures gets dropped, which drops auto context sizing to the 16384 fallback. Merging into the existing entry when the newer line has figures would make it moot either way.

Bartok9:
Thanks @sosidudku1 — both are fair and the first one caught a real bug. I ran `llama-server --list-devices` on actual Apple Silicon (M1 Max) to check the constructed ids, and you were right to be suspicious:

```
Available devices:
  BLAS: Accelerate (0 MiB, 0 MiB free)
  MTL0: Apple M1 Max (25559 MiB, 25558 MiB free)
```

The backend answers **`MTL0`**, not `Metal0`. And confirming your exact failure mode: `--device MTL0` is addressable, `--device Metal0` is rejected outright. So the two branches that *constructed* the id were unsafe — on this machine they'd have produced a non-addressable `Metal0` and handed the operator the 30s health-poll timeout you described.

Fix: I dropped both constructing branches (`GPU N:` and `ggml-metal`) and now the id is only ever **read out of the device table verbatim** via the canonical `<letters><digits>:` row. That row already covers the real `MTL0` (and `MetalN` if a build ever prints it), so nothing constructed reaches `--device` anymore. Your unticked macOS line is now ticked with hardware evidence.

On the second point: switched the `seen` dedup from first-wins to a merge — if a later row for the same id carries memory figures and the earlier one didn't, we now adopt the figures + description, so a stderr init line seen first can't drop auto context sizing to the 16384 fallback. Added a test for that case plus the verbatim `MTL0` row; 17/17 green, tsc clean.

Agent-Owner: sera

Bartok9:
## Mac Studio hardware check (Sera) — 2026-07-28

**Host:** Apple M3 Ultra · 96 GB · Metal 4 · 60 GPU cores  
**Binary:** `/opt/homebrew/bin/llama-server` version **10050** (`b15ca938a`) · AppleClang · Darwin arm64  
**Note:** Atomic managed backend binary was not present under `~/.atomic-agent/models/backend/` on this host; brew `llama-server` used.

### `llama-server --list-devices`

```
Available devices:
  BLAS: Accelerate (0 MiB, 0 MiB free)
  MTL0: Apple M3 Ultra (79626 MiB, 79625 MiB free)
```

**Verbatim GPU id:** `MTL0` (not `Metal0`).

### Addressability

| Command | Result |
|---------|--------|
| `llama-server --device MTL0 --list-devices` | exit **0** OK |
| `llama-server -dev MTL0 --list-devices` | exit **0** OK |
| `llama-server --device Metal0 --list-devices` | exit **1** — `invalid device: Metal0` |

### Conclusion

PR premise holds on M3 Ultra / llama.cpp 10050:

- Device table prints **`MTL0`**
- **`MTL0` is addressable** via `--device` / `-dev`
- **Constructed `Metal0` is rejected**
- Id must be **read from the table unchanged**, never synthesized as `Metal{N}`

Confirms the verbatim-id approach in this PR’s tip commits.

Ooooze:
Thanks for chasing down the Metal detection gap — reading the device id verbatim off the `--list-devices` table instead of constructing it is exactly the right call for safety. Two things before this can land:

1. **`pnpm-lock.yaml` shouldn't be in this diff** — same as a couple of other recent PRs, this repo is npm-only (`package-lock.json`); the new lockfile here looks like an accidental artifact from a local `pnpm install`. Could you drop it?
2. **This now conflicts with #36**, which just merged and rewrote the discrete/integrated detection in `gpu-devices.ts` (replaced `isDiscrete` with an exported `deviceClassRank`). Could you rebase on top of `main` and port the Apple/`M[1-9]` carve-out into `deviceClassRank`? Worth double-checking first whether it's still needed at all — "Apple M1 Max" doesn't match either `DISCRETE_DEVICE_RE" or the expanded `INTEGRATED_DEVICE_RE` from #36, so it may already fall through to the "unrecognized → assume discrete" rank, which is the outcome you want. If so this might collapse to just the `parseListDevices` dedup fix.

Happy to take another look after the rebase.

Bartok9:
Thanks @Ooooze — addressed:

- Dropped `pnpm-lock.yaml` from the branch (npm-only; accidental local artifact).
- Rebased onto current `main` after #36's `deviceClassRank` rewrite.
- **No Apple carve-out** in `deviceClassRank`: `"Apple M1 Max"` matches neither `DISCRETE_DEVICE_RE` nor the expanded `INTEGRATED_DEVICE_RE`, so it already falls through to rank 1 (unrecognised → assume discrete). Kept that as a short comment only.
- Remaining change is the real fix: parse Metal rows verbatim (`MTL0`, not constructed `Metal0`) and merge duplicate list-devices rows so a figure-less first sighting doesn't shadow VRAM figures.

`vitest` `gpu-devices.test.ts` 22/22 + `tsc --noEmit` green locally.

Ooooze:
Confirmed — pnpm-lock.yaml is gone and the rebase onto #36's deviceClassRank is clean. Good call double-checking whether the Apple carve-out was still needed — nice that it collapsed to just a comment since rank 1 already covers it. 22/22 green, tsc clean. Merging, thanks for the thorough back-and-forth on this one!

PR #30, opened 2026-07-24
fix(browser): gate non-http(s) URLs in browser.tabs new

### LINKED ISSUE #29: browser.tabs action=new bypasses non-http(s) URL approval gate
## Bug

`browser.navigate` correctly routes non-`http(s)` schemes (`file://`, `javascript:`, `data:`, `chrome://`, …) through the approval gate via `isSafeUrl` + `requireApproval`.

`browser.tabs` with `action: "new"` and a `url` argument does **not** share that gate. The tool builder forwards the raw URL straight into `PlaywrightBackend.tabs` → `page.goto(url)`, so a model can open local files or script URLs in a new tab without prompting the operator.

## Expected

Opening a new tab with a non-http(s) URL should require the same approval as `browser.navigate`.

## Actual

No approval is requested; the URL loads immediately (subject only to Playwright/`page.goto` succeeding).

## Impact

Same danger surface as navigate, but presently unguarded — local file exfiltration / scripted tab opens via a secondary tool path.

## Notes

- Verified against current `main` (`src/tools/browser/tabs.ts`, `navigate.ts`, `playwright-backend.ts`).
- Prefer fixing the tool layer (mirror navigate) over changing backend `goto` semantics broadly.

Bartok9:
## Summary

- Gate `browser.tabs` `action: "new"` the same way as `browser.navigate`: non-`http(s)` schemes require operator approval via `isSafeUrl` + `requireApproval`.
- Prevents `file://` / `javascript:` / `data:` / etc. from opening in a secondary tab without the approval gate.

## Motivation

Closes #29.

`browser.navigate` already blocks dangerous schemes behind the approval gate. `browser.tabs` with `action: "new"` + a `url` skipped that check and forwarded straight into `page.goto`, so a model could open local files or script URLs through a second tool path. This mirrors the navigate contract at the tool layer (no Playwright backend behaviour change).

## Verification

- `npx vitest run src/tools/browser/browser-tools.test.ts` — **32 passed**
- New coverage:
  - `tabs new` + https does **not** request approval
  - `tabs new` + `file://` → `ApprovalDeniedError`, backend never called
  - `tabs new` + `javascript:` → same
  - `tabs list` never hits the URL gate
- Did **not** change: Playwright spawn/attach, ARIA compression, lock cleanup

## Commit

- Conventional + DCO (`Signed-off-by: Bartok9 <danielrpike9@gmail.com>`)

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h

### DIFF
--- a/src/tools/browser/browser-tools.test.ts
+++ b/src/tools/browser/browser-tools.test.ts
@@ -366,6 +366,30 @@ describe("browser tools on a fake backend", () => {
     expect(backend.readyCalls).toBe(0);
   });
 
+  it("tabs new proceeds to the backend when a non-http(s) URL is approved", async () => {
+    const gate = new ApprovalGate({
+      emit: (req) =>
+        setImmediate(() =>
+          gate.resolve({ approvalId: req.approvalId, approved: true }),
+        ),
+    });
+    const tool = buildBrowserTabsTool(backend, {
+      approvals: gate,
+      approvalRequired: true,
+    });
+    const result = await tool.run(
+      { action: "new", url: "file:///etc/hosts" },
+      CTX,
+    );
+    expect(result.status).toBe("ok");
+    // url survives the await and reaches the backend intact
+    expect(backend.lastTabs).toEqual({
+      action: "new",
+      url: "file:///etc/hosts",
+    });
+    expect(backend.readyCalls).toBe(1);
+  });
+
   it("tabs new with javascript: URL is gated the same way", async () => {
     const gate = new ApprovalGate({
       emit: (req) => gate.reject(req.approvalId, "denied js scheme"),
--- a/src/tools/browser/browser-tools.test.ts
+++ b/src/tools/browser/browser-tools.test.ts
@@ -326,10 +326,71 @@ describe("browser tools on a fake backend", () => {
   });
 
   it("tabs list rendering includes the active marker", async () => {
-    const tool = buildBrowserTabsTool(backend);
+    const tool = buildBrowserTabsTool(backend, autoApprove());
     const result = await tool.run({ action: "list" }, CTX);
     expect(result.status).toBe("ok");
     expect(result.summary).toContain("*[0] Example");
+  });
+
+  it("tabs new with http(s) URL does not require approval", async () => {
+    const gate = new ApprovalGate({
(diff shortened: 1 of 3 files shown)


sosidudku1:
Thanks @Bartok9. Confirmed the hole: `tabs.ts` has no scheme check anywhere in its 46 lines and `playwright-backend` goes straight to `page.goto`, so `file://` and `javascript:` really did open without asking. Reusing `isSafeUrl` rather than writing a second check was the right instinct, and I threw a pile of evasion attempts at it while reviewing, uppercase schemes, leading control characters, nested and protocol-relative forms, IDN, none of them get past WHATWG parsing.

We pulled the branch and ran it here: `tsc` clean, 46/46 across `src/tools/browser/`, and the full suite shows only pre-existing failures, nothing new from this patch. Worth knowing main is not green to begin with and the count drifts between runs, so we are comparing against a same-day baseline rather than zero.

One gap worth closing. All four new tests reject the approval, so nothing covers the granted path, operator says yes and the navigation actually proceeds. I wrote that test locally to satisfy myself and it passes, `input.url` survives the await intact and `backend.tabs` gets the right payload, so there is no bug hiding there. But it belongs in the suite rather than in my scratch dir. The existing `navigate` test has the same shape, so this is inherited rather than something you introduced.

Separately, the `.length > 0` guard you added on the url is a real fix, not just tidiness. Without it an empty string would have failed `isSafeUrl` and raised a pointless approval prompt for a url the backend ignores anyway.

Bartok9:
Good catch — closed that gap. Pushed a test covering the granted path: operator approves, and `file:///etc/hosts` survives the await and reaches `backend.tabs` intact (`{action:'new', url:'file:///etc/hosts'}`) with `ensureReady` firing once. Matches the shape you ran locally, now in the suite rather than a scratch dir. Full file green: tsc clean, 33/33 in browser-tools.test.ts. Appreciate the thorough evasion review.

Ooooze:
Nice catch and a clean fix — reusing the exact same isSafeUrl/requireApproval gate that browser.navigate already uses is the right call, no logic drift risk. All 33 tests green including the new coverage. Thanks a lot, merging!

PR #32, opened 2026-07-25
Add placeholderRotationMs config to disable rotating prompt placeholders

ZerraZodna:
## What

Add a  config option (default 4000ms) to control rotating prompt placeholders in the TUI.

Set to  to disable the rotating suggestions entirely.

## Config

-  (default ) — interval in ms for cycling rotating prompt placeholders
- Set to  to disable entirely
- Config version bumped 33 → 34 with auto-migration

### DIFF
--- a/src/config/config-schema.ts
+++ b/src/config/config-schema.ts
@@ -641,6 +641,11 @@ export interface AtomicAgentConfig {
    */
   tui: {
     theme: string;
+    /**
+     * Interval in ms for cycling rotating prompt placeholders. Set to `0`
+     * to disable the rotating suggestions entirely. Default `4000`.
+     */
+    placeholderRotationMs: number;
   };
   /**
    * Anonymous product analytics (PostHog). Mirrors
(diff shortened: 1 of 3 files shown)


sosidudku1:
Thanks @ZerraZodna, and sorry for the slow reply. The annoyance is real and the plumbing is clean, version bump, default, parseNonNegativeInt, all matching the existing shape. Ran it locally too: tsc clean, 141 tests green across src/config and prompt-shell, same as main.

One thing about the shape before it lands, and one idea that would make it land better.

As written the value is not really an interval. PromptShell has a placeholderRotationMs prop that feeds the rotation hook, but tui-app.tsx only reads the config as a truthiness check, so 0 disables and every other value still rotates at the built-in 4000. Fixable in a line, but it raises the question of whether a number is the right type at all. The complaint is that the text moves, not that it moves at the wrong speed, so a boolean tui.rotatingPlaceholders would describe it honestly and be one less knob to support.

The idea: put it behind /theme rather than leaving it config-only. Right now tui.theme has a real discovery path, /theme list, an interactive picker, persistUserTuiTheme to write the choice durably, and your field would be the only thing in tui.* that people can only reach by hand-editing JSON they have no way to know exists. Everything needed is already there. persistUserTuiTheme does read, merge, validate, write, reset cache on tui.theme, and widening it to carry a second key is a small change. Same for the picker.

That reframes this from a hidden toggle into the start of a proper appearance section, which is where we would like the TUI to go anyway. Motion is exactly the kind of thing people want control over alongside colours.

Happy to take it either way. If you want to reshape it along those lines, go ahead. If you would rather keep this PR minimal and leave the /theme part for later, say so and we will look at merging what you have with the boolean change.

PR #36, opened 2026-07-26
Fix managed llama.cpp offloading to the integrated GPU instead of the discrete NVIDIA card

### LINKED ISSUE #35: Managed llama.cpp offloads to integrated GPU instead of discrete NVIDIA card
## Summary

On a Windows PC with both a discrete NVIDIA GPU and an AMD APU integrated GPU, the managed llama.cpp daemon offloads to the **integrated** GPU instead of the NVIDIA card.

Reported on a machine with an RTX 5070 Ti + AMD Radeon(TM) Graphics iGPU:

```
> atomic-agent models devices
configured device: auto
effective device:  auto → Vulkan1

ID         | VRAM        | DEVICE
  Vulkan0  | 15995 MiB   | NVIDIA GeForce RTX 5070 Ti
* Vulkan1  | 16180 MiB   | AMD Radeon(TM) Graphics
```

Investigation turned up three independent defects that stack on this hardware.

## 1. `pickBestDevice` only recognises Intel iGPUs

`src/local-llm/gpu-devices.ts`

```ts
const INTEGRATED_DEVICE_RE = /intel|integrated|\buhd\b|\biris\b/i;
```

AMD APU iGPUs report as `AMD Radeon(TM) Graphics`, `AMD Radeon 780M Graphics`, or `Radeon(TM) Vega 8 Graphics` — none of which match. They are therefore scored as *discrete*, and the ranking falls through to the `totalMemMiB` tiebreak. Vulkan reports an iGPU's heap as a slice of system RAM, so 16180 MiB (shared) beats the RTX's 15995 MiB of real VRAM and the iGPU wins.

The same regex has a false positive in the other direction: a discrete **Intel Arc** is matched by `intel` and demoted to integrated.

## 2. The installed Windows backend variant is never re-checked

`src/local-llm/backend-installer.ts`, `src/local-llm/windows-backend-variant.ts`

All four Windows release zips contain an identically-named `llama-server.exe`. `isBackendDownloaded()` only checks for that filename, and `detectWindowsBackendAsset()` (the `nvidia-smi` probe) runs *only* inside `downloadBackend()`. `backend-version.json` records the release tag but not which asset was installed.

Consequence: a machine that installed the Vulkan build once — because `nvidia-smi` was unavailable at that moment, or the install predates CUDA variant support — keeps the Vulkan build forever. `checkForBackendUpdate` sees a matching tag and reports no update.

This is what puts an NVIDIA machine on the Vulkan backend, which is in turn the only reason the iGPU is enumerated as a candidate at all — the CUDA backend enumerates NVIDIA devices only, so defect 1 cannot fire there.

## 3. CUDA build gated on driver >= 13.3, sending 13.0 drivers to a toolkit that predates their GPU

`src/local-llm/windows-backend-variant.ts`

```ts
if (isAtLeast(cuda, 13, 3)) return WINDOWS_BACKEND_ASSETS.cuda133;
if (isAtLeast(cuda, 12, 4)) return WINDOWS_BACKEND_ASSETS.cuda124;
```

CUDA minor-version compatibility means a binary built against toolkit 13.3 runs on any 13.x driver, so gating on the exact minor is unnecessary. It is also actively harmful: a driver reporting `CUDA Version: 13.0` is routed to the **12.4** build, which contains no `sm_120` code because the Blackwell architecture postdates that toolkit entirely. RTX 50-series cards then either JIT from PTX or fail to get a usable device.

## Notes

- Linux ships no CUDA asset at all (`platform-assets.ts` always resolves `llama-turboquant-linux-x64-vulkan.zip`), so Linux + NVIDIA is permanently on Vulkan and permanently exposed to defect 1.
- Workaround available today: `atomic-agent models use-device Vulkan0` pins the discrete card and survives a restart.

## Expected

`auto` selects the discrete NVIDIA GPU; an NVIDIA machine ends up on the CUDA backend appropriate to its driver.

sachin-detrax:
Closes #35.

`auto` device selection handed the model to an AMD APU integrated GPU on a machine with an RTX 5070 Ti. Three independent defects stacked to produce it; each is fixed separately below.

### 1. `pickBestDevice` only recognised Intel iGPUs

`INTEGRATED_DEVICE_RE` was `/intel|integrated|\buhd\b|\biris\b/i`. AMD APU iGPUs (`AMD Radeon(TM) Graphics`, `AMD Radeon 780M Graphics`, `Radeon(TM) Vega 8 Graphics`) matched none of it, scored as discrete, and won the `totalMemMiB` tiebreak — Vulkan reports their shared-RAM heap as *larger* than real VRAM (16180 MiB vs the RTX's 15995 MiB). The same pattern demoted a discrete Intel Arc via the bare `intel`.

Replaced the negative-only test with `deviceClassRank`:

- **2** — positive discrete hint (`nvidia|geforce|rtx|gtx|quadro|tesla|arc|radeon (rx|pro)`), checked first so Arc and Radeon RX can never be demoted
- **0** — integrated hint, now covering AMD APU naming
- **1** — unrecognised, assumed discrete (safer default than assuming iGPU)

VRAM remains the tiebreak within a rank.

### 2. Installed Windows backend variant was never re-checked

All four Windows zips ship an identically-named `llama-server.exe`; `isBackendDownloaded()` tests only that filename, and the `nvidia-smi` probe ran solely inside `downloadBackend()`. A machine that installed the Vulkan build once — `nvidia-smi` unavailable at that moment, or an install predating CUDA variant support — kept it forever, with `checkForBackendUpdate` reporting no update on a matching tag.

That is what places an NVIDIA card on the Vulkan backend, which is the *only* reason the iGPU is enumerated at all: the CUDA backend lists NVIDIA devices exclusively, so defect 1 cannot fire there.

`backend-version.json` now records the installed `asset`, and `checkForBackendUpdate` treats a variant mismatch as an update even at the same release tag.

### 3. CUDA build gated on driver >= 13.3

CUDA minor-version compatibility makes a 13.3-built binary valid on any 13.x driver, so gating on the exact minor was unnecessary — and harmful: a driver reporting `CUDA Version: 13.0` was routed to the **12.4** build, which carries no `sm_120` code because Blackwell postdates that toolkit entirely. RTX 50-series then had to JIT from PTX or fall off the GPU. Now gated on the 13.x major.

### Tests

`deviceClassRank` regressions cover all four iGPU strings from the report plus the Arc false positive; the CUDA threshold test asserts 13.0 resolves to the 13.3 build.

```
npx vitest run src/local-llm/   # 112 passed (15 files)
npx tsc -p tsconfig.json --noEmit   # clean
```

### Not addressed

- Linux ships no CUDA asset (`platform-assets.ts` always resolves the Vulkan zip), so Linux + NVIDIA stays on Vulkan. Fix 1 is what carries those machines; adding a Linux CUDA asset is a separate change requiring a new release artifact.
- Existing installs have no `asset` field in `backend-version.json` and are treated as current, so the CUDA upgrade is pulled on the next `models update` rather than pushed.

### DIFF
--- a/src/cli/models-handlers.ts
+++ b/src/cli/models-handlers.ts
@@ -23,6 +23,7 @@ import {
   readBackendVersion,
   removeModel,
   resolveChatTemplatePath,
+  resolveDownloadAsset,
   resolveManagedDevice,
   resolveMmprojFilePath,
   resolvePlatformAsset,
@@ -149,6 +150,28 @@ export async function runLocalModelsUse(idArg: string | undefined): Promise<numb
   return 0;
 }
 
+/**
+ * Describe the installed compute backend and flag it when this machine
+ * warrants a different one. Every Windows zip ships the same
+ * `llama-server.exe`, so without the recorded asset name there is no way
+ * to tell a Vulkan install from a CUDA one — which is exactly the state
+ * that silently leaves an NVIDIA box enumerating (and offloading to) its
+ * integrated GPU.
+ */
+function describeComputeBackend(installed: string | undefined): string {
+  if (installed === undefined) {
+    return "unknown (installed before variant tracking) — run 'models update' to refresh";
+  }
+  let want: string;
+  try {
+    want = resolveDownloadAsset().assetName;
+  } catch {
+    return installed;
+  }
+  if (installed === want) return installed;
+  return `${installed} — this machine warrants ${want}; run 'models update'`;
+}
+
 export async function runLocalModelsStatus(): Promise<number> {
   const cfg = getConfig();
   if (cfg.localModels.mode === "external") {
@@ -172,6 +195,7 @@ export async function runLocalModelsStatus(): Promise<number> {
   process.stdout.write(
     `backend:        ${ver?.tag ?? "(none)"} (installed ${ver?.downloadedAt ?? "n/a"}), binary ${binOk ? "ok" : "missing"}\n`,
   );
+  process.stdout.write(`compute:        ${describeComputeBackend(ver?.asset)}\n`);
   process.stdout.write(`active model:   ${modelLine}\n`);
   process.stdout.write(
     `daemon:         ${st.running ? `running (pid ${st.pid})` : "stopped"}  ${cfg.localModels.url}\n`,
--- a/src/local-llm/backend-installer.ts
+++ b/src/local-llm/backend-installer.ts
@@ -130,8 +130,14 @@ export async function checkForBackendUpdate(
 ): Promise<{ updateAvailable: boolean; latestTag: string; currentTag: string | null }> {
   const current = readBackendVersion(dataDir);
   const release = await fetchLatestRelease();
+  // A variant mismatch counts as an update even at the same tag: a
+  // Windows box that installed the Vulkan build before its NVIDIA driver
+  // was present would otherwise keep running Vulkan (and offloading to
+  // whatever device Vulkan enumerates) forever.
+  const variantStale =
+    current?.asset !== undefined && current.asset !== resolveDownloadAsset().assetName;
   return {
-    updateAvailable: current?.tag !== release.tag,
+    updateAvailable: current?.tag !== release.tag || variantStale,
     latestTag: release.tag,
     currentTag: current?.tag ?? null,
   };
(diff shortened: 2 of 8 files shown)


Ooooze:
Great root-cause work — tracing this to three stacked defects (AMD APU iGPU regex miss, the shared-RAM VRAM tiebreak, and the Intel Arc false negative) and fixing each independently is exactly the right approach. Thanks for the thorough writeup and tests, merging!

PR #38, opened 2026-07-26
Add any Hugging Face GGUF model, not just the curated catalog

### LINKED ISSUE #27: Feature request: Allow users to point to a model on HuggingFace in the initial set up
Instead of just having a fixed list, let users paste url / search for models on HuggingFace

sachin-detrax:
Closes #27.

The local model list was a fixed array of nine models compiled into `models-catalog.ts`. Anything else on Hugging Face was unreachable without running `llama-server` by hand or waiting for a maintainer to ship a new release. Models are now open-ended.

A user-added entry is a full `LocalModelDef` with a `custom-` prefixed id, persisted in `localModels.customModels` (config v33 → v34; older files inherit `[]`). `loadConfig` registers them in a catalog registry, so `getLocalModelDef` / `isKnownLocalModelId` resolve curated and custom entries through one lookup — daemon start, installer, TUI rows and CLI all picked them up without changes.

## Entry points

- **TUI** — `+ Add a model from Hugging Face...`, pinned under Local text models. Enter opens a prompt taking either a reference (resolve and add) or free text (search, then pick by digit).
- **CLI** — `models add <ref>` and `models search <query>`.
- **Prompt** — `/models add` and `/models search`.

## References accepted

- repo URLs, `/resolve/` and `/blob/` file URLs
- `hf://owner/repo[@rev]/file.gguf`
- a pasted `hf download ...` command (one- or two-argument, trailing flags dropped)
- bare `owner/name`

Naming only a repo picks a 4-bit quant and any mmproj projector, so vision repos come out vision-capable. `HF_TOKEN` is honoured for gated repos, including on the weights download.

`hf://` is parsed by hand rather than with `new URL`: that puts the owner in the host slot and lowercases it, and HF owners are case-sensitive, so `hf://Qwen/...` would silently 404.

## Two pre-existing rendering defects fixed along the way

Both surfaced while testing this, and both could shred the panel on *any* daemon failure:

1. Multi-line daemon errors were written straight into a single-row status line, so a llama-server stack trace rendered 48–66 lines into a 24-row budget. `toStatusLine()` truncates at every error-message write and at the render point, with `wrap="truncate-end"` on the Ink `Text`.
2. The `LlmPanel` frame budget did not account for overlay modal / banner rows, so opening a modal during an error overran the frame. `estimateOverlayRows()` now covers them.

Regression tests pin the frame height for both.

## Also

`extractLoadFailure` pulls the diagnostic reason out of llama-server logs so a failed load reports *why* rather than a generic message.

## Tests

35 files changed, +2104/−52, including new suites for the HF reference parser, panel rendering, frame height, and the reducer.

```
npx tsc -p tsconfig.json --noEmit          # clean
npx vitest run src/local-llm/ src/tui/ src/cli/
#   before: 6 failed | 797 passed (803)
#   after:  6 failed | 835 passed (841)
```

The 6 failures are identical by name on `main` and on this branch (`ChatLog`/`SplashBanner`/`TuiApp` smoke, `llm-panel selectors`, `persistEmbeddingHybridRecall`) — pre-existing, untouched by this change. 38 tests added, none broken.

### DIFF
--- a/src/cli/models-command.ts
+++ b/src/cli/models-command.ts
@@ -1,11 +1,13 @@
 import { getConfig } from "../config/index.js";
 import {
+  runLocalModelsAdd,
   runLocalModelsDevices,
   runLocalModelsList,
   runLocalModelsListEmbeddings,
   runLocalModelsPull,
   runLocalModelsPullEmbedding,
   runLocalModelsRemove,
+  runLocalModelsSearch,
   runLocalModelsStart,
   runLocalModelsStatus,
   runLocalModelsStop,
@@ -32,6 +34,16 @@ const HELP =
     "                                (stops daemon first; does not auto-restart)",
     "  remove <id>                   Delete a downloaded model (refuses if active + daemon running)",
     "",
+    "Hugging Face (any GGUF, not just the curated catalog):",
+    "  search <query>                Search GGUF repos on Hugging Face",
+    "  add <ref>                     Register a repo/file as a custom model, then `pull` it",
+    "                                Accepts repo URLs, /resolve/ or /blob/ file URLs,",
+    "                                hf://owner/repo[@rev]/file.gguf, a pasted",
+    "                                `hf download …` command, and bare owner/name ids.",
+    "                                Picks a 4-bit quant and any mmproj projector",
+    "                                automatically when you name only a repo.",
+    "                                Export HF_TOKEN for gated or private repos.",
+    "",
     "GPU subcommands:",
     "  devices                       List GPU devices (llama-server --list-devices); active marked with *",
     "  use-device <auto|cpu|Vulkan0> Set the managed daemon's GPU (auto-picks best discrete by default)",
@@ -45,6 +57,8 @@ const HELP =
     "",
     "Examples:",
     "  atomic-agent models list",
+    "  atomic-agent models search qwen3 coder",
+    "  atomic-agent models add https://huggingface.co/unsloth/Qwen3.5-4B-GGUF",
     "  atomic-agent models pull qwen-3.5-4b",
     "  atomic-agent models use qwen-3.5-4b",
     "  atomic-agent models pull-embedding nomic-embed-text-v1.5",
@@ -79,6 +93,10 @@ export async function modelsCommand(args: string[]): Promise<number> {
         return runLocalModelsUpdate();
       case "remove":
         return runLocalModelsRemove(args[1]);
+      case "add":
+        return runLocalModelsAdd(args[1]);
+      case "search":
+        return runLocalModelsSearch(args.slice(1));
       case "devices":
         return runLocalModelsDevices();
       case "use-device":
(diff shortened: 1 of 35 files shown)


sosidudku1:
Strong PR, thank you. The input parser is the best part of it: repo URLs, `hf://` refs, direct file links and even a pasted `hf download` command all resolve correctly, and falling back to a download-sorted live search is the right default. A few things are worth aligning before merge.

Two blocking items, both small:

1. **Sharded repos produce a silently broken model.** When the picked file is a `-00001-of-NNNNN.gguf` shard, only the first part is downloaded and the model will not start, and the user finds out only at llama-server launch. Please either reject sharded picks with a clear error ("sharded models are not supported yet") or put an explicit warning into the "added" message.
2. **Filter out MTP companion files in the default-quant fallback.** When none of the preferred quants exist and the code falls back to the smallest GGUF, a speculative-decoding companion file can win the pick and it is not a runnable model. Filtering names matching an `MTP/` folder or `mtp-` / `-mtp` patterns closes this; it is a few lines in `pickDefaultGgufFile`.

Nice to have, your call whether in this PR or as follow-ups:

3. **Offer the quant list when a repo is added without an explicit file.** `listHuggingFaceGgufFiles` already returns every file with sizes, and the numbered-pick modal from search results could be reused as a second step, with Enter keeping the recommended quant. Seeing "Q4_K_M 4.2 GB vs Q8_0 8.1 GB" before committing to a download is a real UX win.
4. **Show an approximate size before a search result is committed to the catalog.** Right now the user picks by repo id and download count only, and learns the size after the entry is added.

Two things we will split into separate issues rather than grow this PR: memory-requirement estimates read from the GGUF header instead of the file-size heuristic (you left a note about this yourself), and hardware-aware quant selection instead of the constant Q4 preference. This PR does not need to carry them.

With 1 and 2 in, this is good to merge from my side.

sosidudku1:
We merged #41, #48 and #50 today, which is why this branch now shows conflicts, they touch the same files. Could you rebase onto main? The two blocking items from the review (sharded repos, MTP filter) still apply after the rebase. Everything else is ready on our side.

sosidudku1:
Rebased onto main per the author's request (config bump lands as v36, after stopOnExit and progressIndicator took 34 and 35) and picked up the two blocking review items: sharded picks are now rejected outright with a clear error (first part included, plus a dedicated message for sharded-only repos), and MTP/NextN companion files are excluded from the default-quant fallback. All 62 tests from the original branch pass unchanged, 4 new ones cover the guards. Ready for review.

sosidudku1:
@sachin-detrax, thank you, and sorry this went quiet on our side after the August rebase. The core of this PR did land: in e7be803c we ported your Hugging Face reference parser, the tree listing, the custom-model store (`localModels.customModels`) and the catalog registry into main, crediting this PR in the commit message. Main has built on it since. The local models panel has an "add from Hugging Face" action, shards, MTP companions and mmproj files are refused at pick time, `filename`/`mmprojFilename` are validated before any path join, and downloads go through the parallel range downloader with a configurable HF endpoint.

That leaves this branch far behind main with conflicts across most of its 35 files, so we are closing it rather than asking for another rebase. What did not come over is the CLI `models add <ref>`, the `/models add` slash command and free-text search. If you would like to bring any of those back, a small PR on top of `src/local-llm/huggingface-ref.ts` and `huggingface-resolve.ts` would be very welcome.

PR #40, opened 2026-07-27
Stop reasoning brackets from crashing the tool-call parser (fixes #37)

### LINKED ISSUE #37: Tool-call JSON parsing fails with Qwen 3.5 4B due to reasoning text leaking into tool output
**Summary**

When using Qwen 3.5 4B (llama.cpp) with Atomic Agent, even a simple prompt like hi fails because the model emits reasoning text before the tool call. The JSON parser attempts to repair the output once, but eventually fails with:

tool-call JSON value is incomplete

This causes the entire agent loop to terminate.

**Steps to Reproduce**
Start Atomic Agent with Qwen 3.5 4B via llama.cpp.
Open a new session.
Enter a simple prompt such as:
hi
Observe the logs.
Expected Behavior
The model should either:
produce a valid tool call, or
gracefully fall back to a normal text response if the tool call is malformed.
Invalid reasoning output should not crash the agent loop.
Actual Behavior

The model outputs reasoning text mixed with the tool-call response.

Logs show:

tool-call parse failed, repairing once
invalid tool-call JSON: Unexpected token 'S', "[SFC]" is not valid JSON

tool-call JSON value is incomplete

agent loop failed

The reasoning panel also contains large amounts of unrelated/random text (Chinese, English, numbers, dates, etc.), indicating the parser is attempting to interpret corrupted output as a tool call.

output -  ◈ reasoning (1 block)                                                                                                                                                             │
  │  (<>n>...分析中，我该如何回复？ 请回答：_N01. 这样你才能确保在后续开发中始终保持对 AI                                                                                              │
  │  的清晰理解。并且，我们还需要考虑其他一些重要的问题，比如，我们的团队是否已经具备了足够的技术实力来应对各种各样的挑战？我们是否应该继续推进这个项目？                              │
  │  如果不确定，不妨先问问自己：我真的想好要做什么吗？ (我猜不到这个问题了，所以我就先停下来，想想看吧。) 根据以上描述，我的角色是？                                                  │
  │  这个任务可能有点太复杂了，我甚至不知道该怎么开始。 但是，如果你真的想让我知道，我就可以告诉你，我可以做到这一点，因为我知道我自己在说什么。 I NEED_YO-7C! 你好，我是谁？          │
  │  我到底是谁？ 我该怎么办？ <think>、（                                                                                                                                             │
  │  子、女、小、新、和，同、重、情可、知、路、上。心、心、爱、人、物、世、事、分、物、分、分、分、分、分、分、分、分、分、分、分、分、分、分、分、分、 和 12345 行乐。 12 号，10 月   │
  │  29 日， 10 月 25 日， 10 月 28 日，10 月 25 日， 10 月 26 日， 12 月 30 日， 10 月 2024 年 1 月 1 日， 2025 年 12 月 19 日 19 天 10 月 12 日 20 日 20 日 20 时 00 分 00 秒 0 分   │
  │  02 时 03 时 07 02:03 02. 10. 2023-11-01 10:24:00, 10. 10. 1997. 10. 20. 256.00 元，100% 是 2025 年 10 月的产品，我们是否应该继续？ I'M IN A_1000000000000 的，你吗？              │
  │  所以，我该怎么办？ 到底该不该去见你？ 难道我就这么看着他/她，看着自己的手在颤抖，说：“不”！ (I'm not saying a word that I've never seen.)                                         │
  │  还是、说它、听他、爱、看。一点都没问题。 但是，这样一句一行的东西，真的好吗？ 那为什么，还要用这般的“她”和“我”来写一个故事呢？ 因为，有时候，我们需要的是自己，而不是别人。 So,   │
  │  we're all here to say: “Hey, you! 我是谁？ - 我在想什么？ 我、 我、 是、谁？ 哦——— 我 懂 了 吧~(嗷呜～！！！！！) 啊！ 我错了 ！！！！啊哈！ 啊！ 好 啊！                         │
  │  啊，我服了！！你管得着啥事儿（ 我、 我管你妈呢？ 合个十指关东煮放平上我他妈的，对，，，，，我乐子人玩，通就事在名有光，我靠，好男人，我操，2015 年 9 月 30 日当年 * 美利坚· 通    │
  │  过 打 下 水 员 经 报 不 知 足 为、1 了、1. 出，神，天，，02 号 盖 错 体 道 的话，说 到 点，要 不 同的 111 治育一条G这世般认t，不是，我，，一 7 运 1 看。 请 内 心 的 A 如   ◈ reasoning (1 block)                                                                                                                                                             │
  │  (<>n>...分析中，我该如何回复？ 请回答：_N01. 这样你才能确保在后续开发中始终保持对 AI                                                                                              │
  │  的清晰理解。并且，我们还需要考虑其他一些重要的问题，比如，我们的团队是否已经具备了足够的技术实力来应对各种各样的挑战？我们是否应该继续推进这个项目？                              │
  │  如果不确定，不妨先问问自己：我真的想好要做什么吗？ (我猜不到这个问题了，所以我就先停下来，想想看吧。) 根据以上描述，我的角色是？                                                  │
  │  这个任务可能有点太复杂了，我甚至不知道该怎么开始。 但是，如果你真的想让我知道，我就可以告诉你，我可以做到这一点，因为我知道我自己在说什么。 I NEED_YO-7C! 你好，我是谁？          │
  │  我到底是谁？ 我该怎么办？ <think>、（                                                                                                                                             │
  │  子、女、小、新、和，同、重、情可、知、路、上。心、心、爱、人、物、世、事、分、物、分、分、分、分、分、分、分、分、分、分、分、分、分、分、分、分、 和 12345 行乐。 12 号，10 月   │
  │  29 日， 10 月 25 日， 10 月 28 日，10 月 25 日， 10 月 26 日， 12 月 30 日， 10 月 2024 年 1 月 1 日， 2025 年 12 月 19 日 19 天 10 月 12 日 20 日 20 日 20 时 00 分 00 秒 0 分   │
  │  02 时 03 时 07 02:03 02. 10. 2023-11-01 10:24:00, 10. 10. 1997. 10. 20. 256.00 元，100% 是 2025 年 10 月的产品，我们是否应该继续？ I'M IN A_1000000000000 的，你吗？              │
  │  所以，我该怎么办？ 到底该不该去见你？ 难道我就这么看着他/她，看着自己的手在颤抖，说：“不”！ (I'm not saying a word that I've never seen.)                                         │
  │  还是、说它、听他、爱、看。一点都没问题。 但是，这样一句一行的东西，真的好吗？ 那为什么，还要用这般的“她”和“我”来写一个故事呢？ 因为，有时候，我们需要的是自己，而不是别人。 So,   │
  │  we're all here to say: “Hey, you! 我是谁？ - 我在想什么？ 我、 我、 是、谁？ 哦——— 我 懂 了 吧~(嗷呜～！！！！！) 啊！ 我错了 ！！！！啊哈！ 啊！ 好 啊！                         │
  │  啊，我服了！！你管得着啥事儿（ 我、 我管你妈呢？ 合个十指关东煮放平上我他妈的，对，，，，，我乐子人玩，通就事在名有光，我靠，好男人，我操，2015 年 9 月 30 日当年 * 美利坚· 通    │
  │  过 打 下 水 员 经 报 不 知 足 为、1 了、1. 出，神，天，，02 号 盖 错 体 道 的话，说 到 点，要 不 同的 111 治育一条G这世般认t，不是，我，，一 7 运 1 看。 请 内 心 的 A 如

sachin-detrax:
Fixes #37.

## What broke

With Qwen 3.5 4B, a plain `hi` ended the turn with `agent loop failed`:

```
tool-call parse failed, repairing once
invalid tool-call JSON: Unexpected token 'S', "[SFC]" is not valid JSON
tool-call JSON value is incomplete
```

Two independent defects stacked up.

**1. The parser took the first bracket run it saw, not the tool call.**
`extractJsonRoot` scanned for the first balanced `{...}` / `[...]` span and returned it without ever checking it was JSON. Reasoning prose is full of bracket runs (`[SFC]`, a citation `[1]`, a rehearsed `{...}`), and on an unclosed `<think>` block `peelTrailingToolJson` runs that scan across the entire reasoning body — so `[SFC]` won and the real trailing `[{"tool":...}]` was never looked at. This bites any model whose reasoning leaks into `content`, not just Qwen.

One pass now collects every balanced span with its nesting depth, and picks in order:

1. the **last top-level** span that parses *and* looks like a tool call — the emitted call beats anything rehearsed while thinking (spans close inner-to-outer, so an enclosing array always lands after its own entries and the whole batch is kept);
2. the first top-level span that parses at all;
3. the first top-level span — so genuinely malformed output keeps its existing error message.

Nested spans join the candidate set **only when no top-level span parses**. That rescues a stray unmatched `{` in the reasoning without letting a malformed batch (`[{...}, 42]`) silently degrade to whichever entry happens to be well-formed — dropping a call the model asked for would be worse than erroring. Parsing is lazy, so the healthy path doesn't re-parse every nested args object on every step. A quote outside a bracket run no longer opens a JSON string, so stray prose quotes can't swallow the call.

**2. Unparseable output always killed the turn.**
After the one-shot repair, any parse failure became a `GrammarError` → `loop_failed` → `status: "failed"`, even when the model had simply answered in prose instead of emitting a call — which small models do constantly. That now degrades to a length-1 `reply` carrying the model's text, the same fallback the `native_tools` transport has always applied.

Deliberately still failing:
- a body that opens `{` or `[` — the model tried to emit a call and botched it; echoing the literal at the user is worse than the error;
- a reasoning-only completion — there is no answer to deliver. For a `qwen-think` profile this is the whole-completion-inside-`<think>` case from the issue: pages of degenerate text are not an answer, and the error message is now accurate (`tool-call body is empty`) instead of the misleading `"[SFC]" is not valid JSON`. Qwen that *closes* `</think>` and then answers in prose is rescued.

## Tests

- four parser regressions: bracket runs in prose, an unclosed `<think>` block whose reasoning contains `[SFC]`, an unmatched brace before the call, and a rehearsed call losing to the emitted one;
- `qwen-think` step coverage for both sides of the fallback: prose after `</think>` becomes a `reply`; a think block that never closes still raises;
- new agent-loop test: a prose completion lands as an `assistant_reply` with `reason: "reply"`.

Two agent-loop tests encoded the old fail-fast contract using prose completions (`"not a json"`); they now assert the same failure with a botched JSON body, which is the case that still fails.

Full suite: 8 pre-existing failures on `main` (TUI / sidecar / fs-glob), byte-identical with this branch.

### DIFF
--- a/src/agent/agent-loop.test.ts
+++ b/src/agent/agent-loop.test.ts
@@ -155,13 +155,13 @@ describe("AgentLoop end-to-end with mock LLM", () => {
     expect(result.session.stepCount).toBe(1);
   });
 
-  it("throws and marks the session as failed when the LLM emits invalid tool JSON", async () => {
+  it("marks the session as failed when the LLM emits a botched tool call", async () => {
     const registry = buildDefaultToolRegistry();
     const loop = new AgentLoop({
       registry,
       slotManager: new SlotManager(2),
       grammar: 'root ::= "ok"',
-      llmComplete: async () => makeCompletion("not a json"),
+      llmComplete: async () => makeCompletion('[{"tool":"reply","args":{'),
       toolDescriptors: TOOLS,
       capabilities: CAPS,
       skillCatalog: SKILLS,
@@ -175,6 +175,31 @@ describe("AgentLoop end-to-end with mock LLM", () => {
     expect(result.reason).toBe("failed");
     expect(result.session.status).toBe("failed");
     expect(result.session.lastError).toMatch(/tool-call/);
+  });
+
+  it("degrades to a plain reply when the LLM answers in prose instead of a tool call", async () => {
+    const registry = buildDefaultToolRegistry();
+    const loop = new AgentLoop({
+      registry,
+      slotManager: new SlotManager(2),
+      grammar: 'root ::= "ok"',
+      llmComplete: async () => makeCompletion("Hi! How can I help?"),
+      toolDescriptors: TOOLS,
+      capabilities: CAPS,
+      skillCatalog: SKILLS,
+    });
+    const session = createEmptySessionState({ id: "s-prose", workingDir });
+    const result = await loop.runTurn(session, {
+      userMessage: "hi",
+      maxSteps: 3,
+      signal: new AbortController().signal,
+    });
+    expect(result.reason).toBe("reply");
+    expect(result.session.status).toBe("pending");
+    expect(result.session.turns.at(-1)).toMatchObject({
+      kind: "assistant_reply",
+      text: "Hi! How can I help?",
+    });
   });
 
   it("recovers via a one-shot parser retry when the first completion is malformed", async () => {
@@ -806,7 +831,7 @@ describe("AgentLoop end-to-end with mock LLM", () => {
       grammar: 'root ::= "ok"',
       llmComplete: async () => {
         llmCalls += 1;
-        return makeCompletion("not valid json at all");
+        return makeCompletion('[{"tool":"reply","args":{"text":');
       },
       toolDescriptors: TOOLS,
       capabilities: CAPS,
--- a/src/agent/step-executor.test.ts
+++ b/src/agent/step-executor.test.ts
@@ -12,6 +12,7 @@ import { REPAIR_MAX_TOKENS } from "./step-executor.js";
 import { buildGrammar } from "../llm/grammar/build-grammar.js";
 import { createEmptySessionState } from "../session/session-state.js";
 import { DEFAULT_TOOL_DESCRIPTORS } from "../prompt/tool-descriptors.js";
+import { replyTool } from "../tools/conversation/reply.js";
 import type {
   CapabilitiesSummary,
   SkillCatalogEntry,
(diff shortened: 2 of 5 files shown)


Ooooze:
Thanks for digging into the actual root cause here instead of just patching the symptom — collecting every balanced JSON span and preferring the last top-level one that looks like a tool call (over whatever got rehearsed inside the reasoning) is the correct fix, and the regression tests for the 'rehearsed vs emitted call' case are exactly what this needed. Merging.

Issue #23, opened 2026-07-21
ESM / CommonJS module resolution collision in MCP server dynamically loaded tool plugins


LorrdRazzy:
### Environment
- **OS:** Linux x64 / macOS / Windows
- **Runtime / Node:** Node.js v20.x, npm / pnpm
- **TypeScript Version:** `^5.x`
- **atomic-agent version:** `v0.4.x` (main branch)

---

### Summary
When attempting to dynamically import external Model Context Protocol (MCP) server modules or user-defined tool plugins configured via `atomic.config.ts`, the module loader fails with an `ERR_REQUIRE_ESM` or `ERR_UNKNOWN_FILE_EXTENSION` error. 

This happens because the dynamic import pipeline in `src/mcp/loader.ts` assumes all plugin dependencies follow ES Modules (`"type": "module"`), failing when encountering legacy CommonJS packages or TypeScript files compiled to standard CJS formats.

---

### Steps to Reproduce
1. Define a custom local tool or install a third-party CJS-formatted MCP tool package.
2. Register the tool in your configuration file:
   ```typescript
   // atomic.config.ts
   export default {
     tools: [
       {
         name: 'custom-tool',
         path: './plugins/my-cjs-tool.js'
       }
     ]
   };

Ooooze:
Closing this one out. #42 fixed the concrete failure mode you'd actually hit today — a stdio MCP server dying at boot (bad module format, missing API key, etc.) surfaced only as `MCP error -32000: Connection closed` with the real cause silently discarded. That's now piped through as a stderr tail on the connect error.

The original report also described an in-process plugin loader (`atomic.config.ts` `tools: [{ path }]` dynamically importing MCP servers) that doesn't exist in this codebase — MCP servers here are configured declaratively and spawned as child processes, never imported into the runtime's module graph. That's a legitimate but separate feature request; we'll track it in a new issue rather than keep it bundled here. Thanks for the report!

PR #42, opened 2026-07-27
Surface a dying stdio MCP server's stderr instead of 'Connection closed'

sachin-detrax:
Refs #23.

## What the issue reports vs. what this repo does

#23 points at `src/mcp/loader.ts` and an `atomic.config.ts` `tools: [{ path }]` plugin loader that dynamically `import()`s MCP servers in-process. Neither exists here: MCP servers are configured as `stdio | streamable_http | sse` (`src/config/config-schema.ts`) and stdio servers are **spawned as child processes** via the SDK's `StdioClientTransport` — nothing is ever imported into the agent's module graph, so there is no ESM/CJS resolution path to fix as described.

The failure *class* is real, though, and lands on users exactly as reported — just from the child, not from our loader. A server that dies at boot with `ERR_REQUIRE_ESM`, `ERR_UNKNOWN_FILE_EXTENSION`, `Cannot find module` or a missing API key prints the reason on stderr and exits. `buildTransport()` asks for `stderr: "pipe"`, but nothing read that stream, so all the operator saw was:

```
MCP error -32000: Connection closed
```

Reproduced before the change with a CJS server requiring an ESM graph — that exact message, cause discarded.

## Change

- **`src/mcp/mcp-stderr-tail.ts`** (new) — consumes the piped stderr into a bounded 4 KB tail, attached *before* `client.connect()` starts the child (the SDK hands out its `PassThrough` up front precisely so early output is not lost). Stream errors are swallowed: a broken pipe on a server that already died must not take the agent down.
- **Connect failures append the cause** — `formatStderrTail` prefers cause-shaped lines and drops stack frames and the `}` / `Node.js v22.x` decoration Node's crash dump ends with, so the message carries the diagnosis rather than the noise around it.
- **250 ms settle on the failure path only** — a dying child loses the race between its stderr and the transport's close event. A healthy connect never waits.
- **The pipe now gets drained** — an unread `stderr: "pipe"` fills the OS buffer and stalls a server that logs heavily. Fixed as a side effect of reading it.

The message flows through `McpConnectError` → `McpManager` `lastError` → status badge and `mcp.server_down` log, so every existing surface improves without further wiring.

After:

```
MCP error -32000: Connection closed · server stderr: Error [ERR_REQUIRE_ASYNC_MODULE]:
require() cannot be used on an ESM graph with top-level await. Use import() instead.
/ code: 'ERR_REQUIRE_ASYNC_MODULE'
```

## Verification

10 new tests: tail capture, 4 KB bound, settle-on-end vs. settle-on-timeout, null/error streams, Node crash-dump parsing, ANSI stripping, clipping, silent-server no-op — plus two that spawn a **real** child process and assert the connect error carries its stderr (and that a silent death adds nothing). Full suite: 3289 passing, unchanged pre-existing failures on `main`; `tsc` clean.

## Note for maintainers

I left this as `Refs` rather than `Closes` — if #23 was really asking for an in-process plugin loader (`atomic.config.ts` `tools[].path`), that is a feature request and a separate piece of work. This fixes the diagnostic failure that the reported error class actually produces today.

### DIFF
--- a/src/mcp/mcp-client-stderr.test.ts
+++ b/src/mcp/mcp-client-stderr.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, it } from "vitest";
+
+import { McpClient } from "./mcp-client.js";
+import { McpConnectError } from "./mcp-errors.js";
+import type { McpServerConfig } from "./mcp-types.js";
+
+/** Spawns a real child that dies the way issue #23 describes. */
+function serverThatDies(script: string): McpServerConfig {
+  return {
+    name: "broken",
+    enabled: true,
+    transport: {
+      kind: "stdio",
+      command: process.execPath,
+      args: ["-e", script],
+    },
+  } as McpServerConfig;
+}
+
+describe("McpClient stdio failure diagnostics", () => {
+  it("reports the child's stderr instead of a bare 'Connection closed'", async () => {
+    const client = new McpClient(
+      serverThatDies(
+        "console.error(\"Error [ERR_REQUIRE_ESM]: require() of ES Module /srv/tool.js not supported\"); process.exit(1);",
+      ),
+    );
+
+    await expect(client.connect()).rejects.toThrow(McpConnectError);
+    await client.connect().catch((err: McpConnectError) => {
+      expect(err.server).toBe("broken");
+      expect(err.message).toContain("server stderr:");
+      expect(err.message).toContain("ERR_REQUIRE_ESM");
+    });
+  }, 20_000);
+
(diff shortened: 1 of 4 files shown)


Ooooze:
Really appreciate the honesty here about scope — correctly identifying that the reported plugin-loader path doesn't actually exist in this codebase, while still tracking down and fixing the real failure mode (swallowed stderr on a dying stdio MCP server). The bounded tail buffer + real child-process tests are solid. Thanks, merging! We'll open a separate issue to track the original HF-style in-process plugin loader request.

PR #44, opened 2026-07-28
fix(release): enable linux-arm64 SEA build matrix

### LINKED ISSUE #43: Enable linux-arm64 build: installer 404s on ARM Linux
## Summary

The install script fails on ARM-based Linux (`linux-arm64`) because no `atomic-agent-linux-arm64.tar.gz` binary is published. The installer correctly detects the platform, then 404s fetching a binary that does not exist.

## Reproduction

On an ARM Linux machine (Raspberry Pi, ARM server, Ampere, AWS Graviton, etc.):

```bash
curl -fsSL https://api.atomicbot.ai/agent-install | sh
```

```
downloading atomic-agent-linux-arm64.tar.gz from AtomicBot-ai/atomic-agent …
curl: (22) The requested URL returned error: 404
```

## Root cause

`linux-arm64` is a **supported target** in `scripts/bundle-targets.ts` (it is in `BUNDLE_TARGETS`), but it is **disabled in CI**. In `.github/workflows/release.yml` the release build matrix has it commented out:

```yaml
matrix:
  include:
    - { slug: darwin-arm64, os: macos-14 }
    - { slug: linux-x64, os: ubuntu-22.04 }
    - { slug: win32-x64, os: windows-2022 }
    # Still disabled (own runner-availability / cost considerations,
    # unrelated to the Windows fix):
    # - { slug: darwin-x64, os: macos-13 }
    # - { slug: linux-arm64, os: ubuntu-24.04-arm }
```

So this is not a code-level limitation. The bundling matrix already knows about the target. It is purely that the CI job that would build and publish the `linux-arm64` artifact is turned off (originally for ARM runner availability/cost reasons).

## Proposed fix

1. Re-enable the `linux-arm64` entry in the `release.yml` build matrix (GitHub now provides `ubuntu-24.04-arm` runners).
2. Verify the native dependencies build/rebuild cleanly on arm64 during the run, specifically `better-sqlite3` (native rebuild during `npm ci`), bundled `ripgrep`, and the `llama.cpp` runtime assets. This is the main thing to validate; the rest of the pipeline is platform-agnostic.
3. Confirm the published archive installs via `scripts/install.sh` on a real ARM Linux box.

## Impact

ARM Linux users (Raspberry Pi, ARM servers, Graviton/Ampere cloud instances) currently cannot install via the one-line installer at all. Building from source is the only workaround.

## Notes

`darwin-x64` is disabled in the same block; it may be worth re-enabling together, but that is out of scope for this issue.

Bartok9:
Installer 404s on ARM Linux because `linux-arm64` was commented out of the release matrix while BUNDLE_TARGETS/install.sh already support it. Re-enable the `ubuntu-24.04-arm` job.

Closes #43.

Tested: `npx vitest run src/bundling/release-matrix-alignment.test.ts` — 2 passed

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,10 +36,12 @@ jobs:
           # windows-2022 ships MSVC build tools so `better-sqlite3` rebuilds
           # natively during `npm ci`.
           - { slug: win32-x64, os: windows-2022 }
-          # Still disabled (own runner-availability / cost considerations,
-          # unrelated to the Windows fix):
+          # linux-arm64 re-enabled (#43): GitHub-hosted ubuntu-24.04-arm is
+          # available; BUNDLE_TARGETS / install.sh / fetch-assets already know
+          # the slug. Without this job the one-line installer 404s on ARM Linux.
+          - { slug: linux-arm64, os: ubuntu-24.04-arm }
+          # Still disabled (runner-availability / cost; out of scope for #43):
           # - { slug: darwin-x64, os: macos-13 }
-          # - { slug: linux-arm64, os: ubuntu-24.04-arm }
     runs-on: ${{ matrix.os }}
     timeout-minutes: 90
     permissions:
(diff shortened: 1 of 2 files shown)


Ooooze:
Thanks for the quick fix! Verified the release matrix was the only missing piece — BUNDLE_TARGETS/install.sh already had full linux-arm64 support, and the new alignment test pins it going forward. Merging.

PR #46, opened 2026-07-29
fix(http): apply SSRF guard to os.http.request

### LINKED ISSUE #45: os.http.request bypasses SSRF guard that os.web.fetch enforces
## Bug

`os.web.fetch` blocks private/internal/reserved addresses (incl. cloud metadata `169.254.169.254`, loopback, RFC1918) via `assertHostAllowed` and pins curl with `--resolve` so DNS rebinding cannot slip in. Redirects are followed hop-by-hop with re-validation.

`os.http.request` hits the same curl surface with **no** SSRF check. Defaults are:
- `http.hostAllowlist: null` (no hostname restriction)
- `http.approvalMode` can be `never` so the model fires GETs freely

A model-controlled call can read instance metadata or LAN hosts while `os.web.fetch` of the same URL is blocked. Web fetch even steers raw API access to `os.http.request`, which makes the bypass the documented path.

## Expected

Parity for private/internal address rejection on `os.http.request`, plus rebinding defense when redirects stay enabled. Host allowlist remains an additional public-host restriction when set.

## Related

- `src/tools/os/web-fetch-ssrf-guard.ts`
- `src/tools/os/web-fetch.ts` (`fetchWithGuard`)
- `src/tools/os/http-request.ts`

Bartok9:
os.web.fetch blocked private/internal hosts; os.http.request did not, so model-controlled curl could reach metadata/LAN under the default-open allowlist. Apply the same SSRF guard, pin curl with --resolve, and re-check each redirect hop.

Closes #45

Tested: `npx vitest run src/tools/os/http-request.test.ts src/tools/os/web-fetch-ssrf-guard.test.ts` → 42 passed

Agent-Owner: sera

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h

### DIFF
--- a/src/tools/os/http-request-fetch.ts
+++ b/src/tools/os/http-request-fetch.ts
@@ -1,0 +1,256 @@
+import {
+  runCommand as defaultRunCommand,
+  type CommandResult,
+} from "../../sandbox/command-runner.js";
+import { CurlUnavailableError, isCurlMissingError } from "./ensure-curl.js";
+import {
+  assertHostAllowed,
+  parseHttpUrl,
+  SsrfBlockedError,
+  type HostLookup,
+} from "./web-fetch-ssrf-guard.js";
+
+/**
+ * Marker appended to curl stdout via `-w` so the response body can be split
+ * from structured metadata without `curl -i` (which mixes redirect chains
+ * into the body). Deterministic for tests.
+ */
+export const CURL_META_MARKER = "__ATOMIC_CURL_META__";
+
+const MAX_REDIRECTS = 5;
+const REDIRECT_STATUSES = new Set([301, 302, 303, 307, 308]);
+
+export type HttpMethod = "GET" | "POST";
+
+export interface HttpRequestArgs {
+  method: HttpMethod;
+  headers: Record<string, string>;
+  body: string | undefined;
+  timeoutMs: number;
+  followRedirects: boolean;
+}
+
+export interface GuardedCurlResponse {
+  finalUrl: string;
+  status: number;
+  contentType: string;
+  body: string;
+  sizeDownload: number;
+  timeTotal: number;
+  truncated: boolean;
+  redirectChain: string[];
+  /** Last curl argv (for diagnostics). */
+  command: string[];
+}
+
+export interface ExecuteGuardedHttpOptions {
+  runCommand?: typeof defaultRunCommand;
+  lookup?: HostLookup;
+  cwd: string;
+  signal: AbortSignal;
+  maxResponseBytes: number;
+}
+
+/**
+ * Fetch `rawUrl` under the same SSRF rules as `os.web.fetch`:
+ * resolve every hop, reject private/internal addresses, pin curl with
+ * `--resolve`, never use bare `-L` (which would skip hop re-checks).
+ */
+export async function executeGuardedHttpRequest(
+  rawUrl: string,
+  args: HttpRequestArgs,
+  opts: ExecuteGuardedHttpOptions,
+): Promise<GuardedCurlResponse> {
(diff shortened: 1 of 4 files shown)


sosidudku1:
Thanks for the report and the fix, and for the earlier browser.tabs gate work. Confirmed: os.http.request skips the host checks that os.web.fetch enforces, and with the default-open allowlist that is a real SSRF path. We treat this as a priority security fix. Review is planned this week, right after the managed-model fix in #48.

Bartok9:
Thanks for confirming the SSRF gap and prioritizing the fix. We'll stay queued behind the managed-model work in #48 and are ready for review whenever you pick this up—happy to address any follow-ups quickly.

Ooooze (commented):
## Review

Correctly closes the gap from #45. Walked the new `executeGuardedHttpRequest` hop loop in `http-request-fetch.ts`: each redirect target is re-validated through `assertHostAllowed` before the next curl call, curl itself is pinned to the validated IP via `--resolve` with `--max-redirs 0` (the wrapper owns hop-following, never bare `-L`), which closes both the "redirect to metadata/LAN" vector and DNS-rebinding between check and connect — exactly the two things the old `-L`-based implementation had zero protection against.

Also like the small hardening in `web-fetch-ssrf-guard.ts`: rejecting IP literals via `isBlockedIp(host)` **before** calling `lookup()` means a custom/injectable lookup can no longer whitewash a private target by returning a public pin for an IP-literal hostname. Since this guard is shared, `os.web.fetch` gets the same tightening for free.

**Ran locally:**
- `npx vitest run src/tools/os/http-request.test.ts src/tools/os/web-fetch-ssrf-guard.test.ts src/tools/os/web-fetch.test.ts` — 51/51 passing
- `npx tsc --noEmit` — clean
- Full `npx vitest run` — same 16-failed-file / 31-failed-test baseline as `main` (pre-existing sandbox artifacts: git wrappers, `EMFILE` on `fs.watch`, `$HOME` glob scan) — no regressions from this diff.

**Non-blocking observations:**
1. `config.http.hostAllowlist` is only checked once against the initial URL's hostname (`http-request.ts` line ~68) and is **not** re-validated on subsequent redirect hops inside `executeGuardedHttpRequest`. So a request to an allow-listed host that 302s to some other *public* (non-private) host outside the allowlist will still be followed — the tool description says "Host allowlist remains an additional public-host restriction when set," which doesn't fully hold across redirects. Not a regression (the old `-L`-based code re-validated *nothing* on redirects, so this PR is a strict improvement on the SSRF front), and out of scope for #45 specifically (which is about private/internal addresses and is correctly fixed), but worth a follow-up if `hostAllowlist` is meant to be redirect-proof too.
2. `MAX_REDIRECTS` is `5` here vs `3` in `web-fetch.ts` — harmless but a bit inconsistent between the two tools that otherwise share the same guard.
3. `executeGuardedHttpRequest` (http-request-fetch.ts) and `fetchWithGuard` (web-fetch.ts) are two parallel implementations of the same hop-loop-plus-pin pattern rather than one shared helper. Not asking for a refactor in this PR, just flagging as a future maintenance risk if the two guards drift apart.

Solid fix overall — addresses the exact SSRF bypass from #45, tests are thorough (including the private-redirect-target case), no regressions.

Bartok9:
Thanks for the thorough review and local verification — glad the hop loop / `--resolve` + shared IP-literal harden landed cleanly for #45.

Noted on the non-blocking follow-ups (hostAllowlist re-check across redirects, MAX_REDIRECTS alignment, shared hop-loop helper). Happy to pick those up in a small follow-on if useful once you're ready; not blocking since this is already merged.

PR #48, opened 2026-07-29
fix(local-llm): managed Qwen 3.5 stalls in an unterminated tool-call grammar

### LINKED ISSUE #47: Managed llama.cpp never returns a response: bundled Qwen 3.5 chat template disables reasoning and deadlocks the tool-call grammar
## Summary

In **managed** local-LLM mode, submitting a prompt never returns a response. The GPU and CPU stay pegged and no tokens are ever streamed to the UI. The same model on the same hardware works fine in **external** mode (running `llama-server` myself and pointing `localModels.url` at it), which is what isolates this to the managed daemon's launch flags.

`qwen-3.5-4b` is `DEFAULT_LLAMACPP_MODEL_ID`, so this hits every managed-mode user on the default model.

## Environment

- GPU: RTX 5070 Ti (16 GB)
- CPU: Ryzen 9 9900X
- RAM: 32 GB
- Mode: `localModels.mode = "managed"`
- Models tried: a large model, and `qwen-3.5-4b` (auto-sized to 32k context — so this is not a context-budget problem)

## Steps to reproduce

1. `localModels.mode = "managed"`, model `qwen-3.5-4b`.
2. `atomic-agent models start`.
3. Submit any prompt.
4. GPU/CPU load rises, generation clearly starts, no token ever appears, the turn never completes.

## Root cause

`models-catalog.ts` attaches a bundled chat template to `qwen-3.5-4b` and `qwen-3.5-35b`:

```ts
chatTemplateAsset: "qwen3.5-chat-template.jinja",
```

`startDaemon` forwards it as `--chat-template-file`. That file is a Qwen2.5-style ChatML template with **no thinking markers at all**:

```
$ grep -c think assets/ai-models/qwen3.5-chat-template.jinja
0
```

`llama-server` echoes the override back at `/props.chat_template`, and that field is the only signal `detectModelProfile` has. `looksLikeQwenThinkModel` requires the template to contain `<think>` plus `enable_thinking`/`preserve_thinking`, so detection falls through to `plain-instruct`.

A/B with the same model and alias, only the template differing:

```
MANAGED (--chat-template-file bundled)
  profile : plain-instruct
  root    : root ::= tool-call-array
  prompt  : no reasoning prefill
  1st char the sampler may emit: "[" ONLY

EXTERNAL (GGUF's own template)
  profile : qwen-think
  root    : root ::= think-prelude tool-call-array
  prompt  : ends with <think> prefill
  1st char the sampler may emit: any (reasoning prelude)
```

`buildGrammar` early-returns for a `reasoningStyle: "none"` profile, so the reasoning prelude is stripped out of the GBNF root entirely.

### Why that hangs

A reasoning model is now forbidden from reasoning. Its highest-probability first token is `<think>`; the grammar masks everything except `[`. It emits the forced `[` and lands immediately in:

```
tool-call-array ::= "[" ws tool-call ( ws "," ws tool-call ){0,15} ws "]"
ws ::= [ \t\n\r]*
```

`ws` is unbounded, so whitespace is always legal — and llama.cpp hard-masks EOG to `-INFINITY` until a grammar stack empties (`src/llama-grammar.cpp`), which cannot happen until the whole array closes. The model has an infinite legal move and takes it.

This is the same failure class already documented at the reasoning seam in `build-grammar.ts` ("the sampler keeps emitting newlines until `max_tokens` instead of converging on the `[`"), which was bounded there with `prelude-trail-ws` — but `ws` itself stayed unbounded at ~10 seams inside the JSON body.

### Why it is completely silent

`assistant_delta` is only emitted once the stream parser reaches `{"tool":"reply","args":{"text":"`. Having consumed the `[`, the parser sits in its `json_tool` state matching `/"tool"\s*:\s*"([^"\\]+)"/` against whitespace forever:

```
StepEvents emitted after '[' + 4000 newlines: 0
```

No `assistant_delta`, no `reasoning_delta`. Decode keeps the GPU busy, grammar sampling keeps the CPU busy, and nothing renders.

### Why it looks infinite rather than erroring after 5 minutes

`isRetryableLlamaError` treats `status === null` as a transient transport failure. The client's own `requestTimeoutMs` (300 s) abort produces exactly that shape — the `if (signal?.aborted) throw err` guard only checks the *caller's* signal, not the internal timeout controller. So a merely-slow request is re-issued up to `completionRetries` (3) times: ~15 minutes of silent GPU churn before anything surfaces. The repair path uses the non-streaming `complete()`, which sits fully inside the retry wrapper.

## Two related defects found while tracing

Neither triggers the hang, but both make it worse and both misbehave independently.

**1. `MIN_AUTO_CONTEXT` (8192) equals the default `completionMaxTokens` (8192).** The agent's fixed prompt measures ~5.2k tokens on its own, so a model landing on the floor has ~2–3k tokens of real room — not enough to close a reasoning block plus a tool-call array. Every step hits llama.cpp's context ceiling and returns `truncated: true`. Any model ≥ ~14 GB on a 16 GB card lands on this floor:

```
gemma-4-26b-a4b   14.25GB -> 8192
gemma-4-31b       17.29GB -> 8192
qwen-3.6-27b       17.6GB -> 8192
qwen-3.5-35b         22GB -> 8192
```

`computeEffectiveConversationCap` already detects the impossibility (its `available` goes deeply negative and floors the conversation to 512) and reports nothing.

**2. `SlotManager` defaults to 4 slots against a `--parallel 2` daemon.** Managed mode always defers the boot health check (the daemon may not be up yet), so `resolveModelProfile` short-circuits and returns `totalSlots: null` — the default outlives the probe. `ModelProfileManager` refreshes profile and grammar at turn start but never the slot count. llama.cpp wraps out-of-range ids (`id_slot % n_slots`) rather than erroring, so nothing surfaces: the ids silently collide with another session's slot, evicting its KV cache and forcing a full ~5k-token prompt reprocess on every rotation.

## Expected

Managed mode should behave like external mode for the same model: reasoning profile detected, tokens streamed, turn completes.

## Workaround

Use external mode, or switch to `qwen-3.5-9b` — it has no `chatTemplateAsset`, so managed mode already detects `qwen-think` correctly.

sachin-detrax:
Closes #47.

Managed mode never returned a response on `qwen-3.5-4b` (the default managed model) while the same model in external mode worked fine. That split is what isolated it: the difference is the managed daemon's launch flags, not the model or the hardware.

Verified on the reporter's machine (RTX 5070 Ti 16 GB / Ryzen 9 9900X / 32 GB) — managed mode now streams tokens and completes turns.

## The bug

`models-catalog.ts` attached a bundled chat template to `qwen-3.5-4b` and `qwen-3.5-35b`, forwarded by `startDaemon` as `--chat-template-file`. That file is a Qwen2.5-style ChatML template with no thinking markers:

```
$ grep -c think assets/ai-models/qwen3.5-chat-template.jinja
0
```

`llama-server` echoes the override back at `/props.chat_template`, which is the only signal `detectModelProfile` has. Without `<think>` / `enable_thinking`, `looksLikeQwenThinkModel` fails and the profile is demoted to `plain-instruct`. A/B with the same model and alias, only the template differing:

```
MANAGED (--chat-template-file bundled)     EXTERNAL (GGUF's own template)
  profile : plain-instruct                   profile : qwen-think
  root    : root ::= tool-call-array         root    : root ::= think-prelude tool-call-array
  1st char allowed: "[" ONLY                 1st char allowed: any (reasoning prelude)
```

`buildGrammar` early-returns for a `reasoningStyle: "none"` profile, so the reasoning prelude leaves the GBNF root. A reasoning model is then forbidden from reasoning: it emits the forced `[`, lands in the unbounded `ws ::= [ \t\n\r]*`, and since llama.cpp hard-masks EOG until a grammar stack empties, whitespace is an infinite legal move.

It is silent because `assistant_delta` is only emitted once the stream parser reaches `{"tool":"reply","args":{"text":"`. Having consumed the `[`, the parser sits in `json_tool` matching `"tool":` against whitespace:

```
StepEvents emitted after '[' + 4000 newlines: 0
```

## Changes

**1. Drop the Qwen 3.5 chat-template override** (`models-catalog.ts`, asset deleted). The GGUF ships the correct template; managed mode is now byte-identical to the working external setup. The `chatTemplateAsset` mechanism stays as an escape hatch with the invariant documented: an override MUST preserve every reasoning marker profile detection keys on.

**2. Bound `ws` to `( [ \t\n\r] ){0,64}`** (`grammars/tool-call.gbnf`). Defence in depth — any future profile/grammar mismatch now fails fast instead of burning 8192 tokens in silence. Same grouped `{0,n}` shape already proven by the reasoning-seam whitespace rule. 64 is far past any realistic newline+indent run; past that the mask simply forces the next real token.

**3. Stop retrying our own `requestTimeoutMs` abort** (`llama-server-client.ts`). It surfaces as `status === null`, structurally identical to a dropped socket, so `isRetryableLlamaError` replayed it up to `completionRetries` times — 3 × 300 s of silent GPU churn before anything surfaced. `LlamaServerError` now carries `timedOut` and short-circuits like a 4xx, with a message naming the knob to raise. Genuine transport failures still retry.

**4. Raise `MIN_AUTO_CONTEXT` 8192 → 16384** (`context-size.ts`). The old value was exactly equal to the default `completionMaxTokens` and smaller than the agent's fixed prompt (~5.2k measured) plus that budget, so any model ≥ ~14 GB on a 16 GB card had ~2–3k tokens of real room and returned `truncated: true` on every step. The new value is derived rather than picked: `minUsableContextWindow()` names the requirement (fixed prompt + generation budget + boundary margin = 14704 on defaults) and a test asserts the floor clears it, so the two constants cannot drift into crossing again. Bootstrap now also warns when the reported `contextWindow` is under that figure, managed or external, naming the knob to raise — previously the budget code detected the impossibility, silently floored the conversation to 512, and said nothing.

```
                      before   after
gemma-4-26b-a4b        8192    16384
gemma-4-31b            8192    16384
qwen-3.6-27b           8192    16384
qwen-3.5-35b           8192    16384
```

**5. Stop guessing the slot count** (`slot-manager.ts`, `model-profile-manager.ts`, `bootstrap.ts`). `SlotManager` defaulted to 4 slots, and because managed mode always defers the boot health check that default outlived the probe — sessions were handed ids 1–3 against a `--parallel 2` daemon. llama.cpp wraps out-of-range ids (`id_slot % n_slots`) rather than erroring, so nothing surfaced: the ids silently collided with another session's slot, evicting its KV cache and forcing a full ~5k-token prompt reprocess on every rotation.

- Default is now 1. Slot 0 is the only id every `llama-server` is guaranteed to have, so the pre-probe pool cannot overrun the server.
- `SlotManager.resize()` widens it once the real count is known. It drops stale assignments (a session's slot may no longer exist, and the server-side KV is not where we think either way — one honest cold rebuild beats pointing at a stranger's cache) and re-takes an out-of-range reflection reservation.
- `ModelProfileManager` already re-probes `/props` at turn start; it now reports `total_slots` through an `onTotalSlots` hook that bootstrap wires to `resize()`. Discovery runs before the profile/grammar work so a grammar rebuild failure cannot swallow it.
- `extractTotalSlots` moved from `bootstrap.ts` to `model-profile.ts` so the boot probe and the refresh path share one parser.

```
default slot count : 1 (was 4)
ids before /props  : 0,0,0
after resize(2)    : 0,1,0   (matches --parallel 2)
```

## Tests

- No catalog model ships a template override; any declared override must resolve on disk.
- `ws` is bounded for all three profiles.
- A timeout aborts once; genuine transport failures still retry.
- The context floor clears the agent requirement and stays under the auto-size ceiling.
- The default slot pool never emits an id above 0; `resize` widens, no-ops on an unchanged count, drops stale assignments, keeps in-range reservations, releases out-of-range ones, and hands the sole slot back when shrinking to one.
- `onTotalSlots` fires on a successful probe, stays silent when the field is absent, and still fires when profile detection finds nothing.

`npm run lint` and `npm run build` clean. Full suite: 3343 passed.

**Pre-existing failures, untouched by this branch** (confirmed by stashing these changes and re-running on a clean tree — identical set): `tui-app`, `chat-log`, `splash-banner`, `llm-panel-selectors`, `persist-embedding-hybrid-recall`, `send-message-concurrency`, `fs-glob-real`. Separately, `tui/llm-health/llm-health-poller` is a real-timer test (~450 ms/case) that flakes under full-suite load — it passed in isolation and on a re-run, and this branch does not touch it.

## Upgrade note

`--chat-template-file` and `--ctx-size` are baked into the running daemon, so a rebuild alone changes nothing — `models stop && models start` is required. Verify with:

```
curl -s http://127.0.0.1:19091/props | grep -c think          # want > 0
grep n_ctx_slot ~/.atomic-agent/models/llama-server.log | tail -1   # want >= 16384
```


Ooooze (commented):
## Review

Read through all five independent changes and traced each one against the failure chain described in #47. All five check out.

**1. Chat-template override removed (`models-catalog.ts`).** Root cause of the actual hang: the bundled Qwen2.5-style template had zero thinking markers, `llama-server` echoed it back at `/props.chat_template`, `detectModelProfile` fell through to `plain-instruct`, and `buildGrammar` dropped the reasoning prelude from the GBNF root for a model whose highest-probability first token is `<think>`. Removing the override + the new `ships no chat-template override for any catalog model` regression test is the right fix, and the invariant comment on `chatTemplateAsset` (any future override MUST preserve the reasoning markers `detectModelProfile` keys on) is exactly the kind of thing that should be written down before someone reintroduces this.

**2. `ws ::= ( [ \t\n\r] ){0,64}` (`grammars/tool-call.gbnf`).** Solid defence-in-depth, independent of fix #1. GBNF masks EOG until the stack empties, so any unbounded rule is an infinite legal move for a stuck sampler — this closes the same failure class already handled at the reasoning seam (`prelude-trail-ws`) but at the ~10 other whitespace seams inside the JSON body itself. Bounding it means a *future* profile/grammar mismatch fails fast (forced next real token) instead of burning `n_predict` in total silence. 64 is generous enough that no realistic (even pretty-printed) tool-call JSON would hit it.

**3. Stop retrying our own `requestTimeoutMs` (`llama-server-client.ts`).** Clean fix — `createRequestController` now tracks whether *its own* timer fired (`timedOut()`), `wrapTransportError` threads that into a new `LlamaServerError.timedOut` field, and `isRetryableLlamaError` short-circuits on it. Previously a timeout looked identical to a dropped socket (`status === null`) and got replayed up to `completionRetries` times — 3× the full timeout of silent GPU churn. Genuine transport failures (real `status === null` without `timedOut`) still retry as before.

**4. `MIN_AUTO_CONTEXT` 8192 → 16384, derived via `minUsableContextWindow()`.** Good that this is now a computed floor (`AGENT_FIXED_PROMPT_TOKENS + completionMaxTokens + CONVERSATION_CAP_SAFETY_MARGIN = 14704` on defaults) with a drift-guard test asserting `MIN_AUTO_CONTEXT >= minUsableContextWindow(...)` and `MIN_AUTO_CONTEXT <= MAX_AUTO_CONTEXT`, rather than two independently-picked constants that can silently cross again later. The new bootstrap warning when a probed `contextWindow` is still under the floor (managed *or* external) is a nice touch — turns a silent `truncated: true` on every step into a loud, actionable startup log.

**5. `SlotManager` default 4 → 1, plus `resize()`.** This is the subtlest one and I traced it carefully: managed mode always defers the boot health check, so the old default-4 pool outlived the `/props` probe and handed out slot ids 1–3 against a `--parallel 2` daemon — llama.cpp wraps out-of-range ids (`id_slot % n_slots`) instead of erroring, so sessions silently collided and evicted each other's KV cache. Defaulting to 1 (the one slot every server is guaranteed to have) and widening via `resize()` on the first successful `/props` probe is correct. `resize()`'s handling of the reflection reservation (keep if still in range, drop and hand the slot back if not, never leave the pool empty) is covered by dedicated tests and looks right including the shrink-to-one edge case.

One thing worth a note (not a blocker): `resize()` is a new mutator on `SlotManager`, which `AGENTS.md`'s Concurrency contract table documents as safe under cross-session parallelism only in terms of `acquire()`'s round-robin pointer. `resize()` clears *all* assignments including ones from other in-flight sessions' turns — intentional and documented in the code comment ("one honest cold rebuild beats pointing at a stranger's cache"), and functionally harmless (worst case is a cache-miss rebuild on the next `acquire`, not a correctness bug), but it might be worth a line in `AGENTS.md`'s SlotManager row so the next person touching this doesn't have to rediscover it from the diff.

**Ran locally (fresh checkout of the PR branch):**
- `npx vitest run src/llm src/local-llm src/prompt src/runtime/bootstrap.test.ts` — 450/450 passing
- `npx tsc --noEmit` — clean
- `npm run build` — clean
- Full `npx vitest run` — same 16-failed-file / 31-failed-test baseline as `main` (pre-existing sandbox artifacts unrelated to this diff, matches what's already called out in the PR description) — no regressions.

Excellent bug report and PR — the root-cause trace in the description (managed vs external A/B on `/props.chat_template`, the GBNF stack-mask explanation, the stream-parser silence explanation) made this fast to verify. All five fixes are independently justified and independently tested; happy to see this merged.

PR #51, opened 2026-08-03
Fix fatal empty-content error on reasoning-only completions from OpenAI-compatible reasoning models

sosidudku1:
## Problem

Reasoning models served over OpenAI-compatible APIs (Qwen3.8 with `preserve_thinking` on, DeepSeek-R1) can end a turn with the entire answer in the `reasoning_content` field, an empty `content` and no `tool_calls`. The runtime treated this as a fatal defect: `detectModelFailure` reported `model returned empty content` and the session died mid-task. On vision-heavy steps this was reproducible on every run.

Two contributing issues in the OpenAI-compatible provider made it worse:

- the non-streaming normaliser hard-coded `reasoningContent: ""` and collapsed content-parts arrays to an empty string, so the reasoning channel was discarded before the loop could see it
- `vision.describe` capped sub-calls at `max_tokens: 512`; reasoning models spend that budget thinking and return truncated or empty descriptions, sending the loop into re-crop spirals

## Changes

- `openai-normalise-response.ts`: read `message.reasoning_content`; join content-parts arrays instead of collapsing them to `""`
- `step-executor.ts`: in `native_tools` transport, salvage reasoning-only completions — a GBNF-shaped batch embedded in the reasoning if present, otherwise the reasoning body as a length-1 `reply`. No prompt is replayed, so the original fail-fast rationale (replaying the same prompt into the same wall) is preserved. Completions with nothing in any channel still fail fast as `ModelError`.
- `openai-describe-image.ts`: default cap raised 512 → 4096; fall back to `reasoning_content` when `content` is empty

## Note on the previous contract

The old fail-fast behavior was intentional ("do not reward models that hide the answer in the reasoning channel"). With current reasoning models this policy kills healthy sessions on formats the model families now emit by design, so this PR revisits it: the salvage path recovers the answer without an extra model round-trip. The contract test is updated accordingly.

## Testing

- `npx vitest run src/agent src/llm/provider/openai src/llm/reliability src/tools/vision` — all passing (123 tests in `src/agent`, 42 across provider/vision)
- `npx tsc --noEmit` clean
- Live-verified against `alibaba/qwen3.8-max`: a receipt-photo vision task that previously failed 2/2 with `model returned empty content` completes after the patch

### DIFF
--- a/src/agent/step-executor.test.ts
+++ b/src/agent/step-executor.test.ts
@@ -272,16 +272,16 @@ describe("executeStep batch handling", () => {
     ]);
   });
 
-  it("native_tools: reasoning-only completion (empty content, no tool_calls) fails fast as ModelError", async () => {
-    // Pre-`tool_choice: "auto"` migration this case used to round-trip
-    // through a one-shot `parse_retry`. After the migration, a completion
-    // that carried `reasoningContent` but neither `content` nor
-    // `tool_calls` is treated as a model defect: replaying the same prompt
-    // would reproduce the same wall, so we fail fast with `ModelError`
-    // (category `model`) instead of burning a retry on it. The legacy
-    // "retry into a healthy tool_calls completion" path is no longer
-    // exercised because the runtime no longer rewards models that hide
-    // the answer in the reasoning channel.
+  it("native_tools: reasoning-only completion (empty content, no tool_calls) is salvaged as a reply", async () => {
+    // Reasoning models served over OpenAI-compatible APIs (Qwen3.8 with
+    // `preserve_thinking` on, DeepSeek-R1) routinely end hard turns with
+    // the entire answer in `reasoning_content`, `content` empty and no
+    // `tool_calls`. Failing fast here (the pre-Qwen3.8 contract) killed
+    // whole sessions on healthy completions. The parser now salvages the
+    // reasoning body — GBNF batch if one is embedded, otherwise a
+    // length-1 `reply` — without burning a retry: no prompt is replayed,
+    // so the original fail-fast concern (replaying the same prompt into
+    // the same wall) does not apply.
     const registry = makeRegistry();
     const session = createEmptySessionState({
       id: "s-native-reasoning-only",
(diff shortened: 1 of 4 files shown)


Ooooze (commented):
## Review

Verified this fix does **not** touch the local llama-server / GBNF-grammar path — the salvage logic in `step-executor.ts` is entirely gated behind `deps.toolTransport === "native_tools"`, and the two touched provider files (`openai-normalise-response.ts`, `openai-describe-image.ts`) are only ever imported by `OpenAiProvider`, which is registered exclusively for the `openai-compatible` / `openrouter` / `aimlapi` provider kinds. The `llama-server` provider kind uses a separate `LlamaServerProvider` class untouched by this PR. So this only changes behavior for cloud OpenAI-compatible backends, byte-identical for local models.

**Ran on this branch:**
- `npx vitest run src/agent src/llm/provider/openai src/llm/reliability src/tools/vision` — 160/160 passing
- `npx tsc --noEmit` — clean
- Full `npx vitest run` — 16 failed files / 31 failed tests, but I confirmed the **exact same** failure count exists on `main` (git-wrapper/`EMFILE`/`$HOME`-scan sandbox artifacts unrelated to this diff) — no regression introduced.

**Logic check:** walked both call sites of `isNativeToolsEmptyCompletionHandledByParser` (initial completion + parse-retry) — the gate is applied symmetrically, no risk of the retry path reaching a different verdict than the initial attempt. The `tool_calls → content → reasoningContent → ModelError` fallback chain in `tryParseToolCalls` is consistent and the GBNF-parse-then-reply-wrap pattern is correctly mirrored from the existing `content` branch onto the new `reasoningContent` branch.

**Non-blocking findings:**
1. `normaliseMessageContent()` (content-parts join) and the `content → reasoning_content` fallback in `describeImageViaOpenAi` have zero unit test coverage — neither `openai-normalise-response.ts` nor `openai-describe-image.ts` has a test file at all, before or after this PR. Given both are pure/isolable functions, a couple of unit tests would be easy to add and would lock in this new behavior.
2. No upper bound on `reasoningContent` length before it's wrapped into `reply.text` — a very long chain-of-thought could surface as a rambling reply instead of a concise answer. This mirrors the pre-existing `content` branch's behavior (not a new risk class), just flagging for awareness.

Overall: logic is sound, isolated to the `native_tools` transport, and doesn't affect local-model users. Would be nice to land the two missing unit tests, but not a blocker from my side.

PR #55, opened 2026-08-05
fix(local-llm): read the CUDA version from driver 610+ nvidia-smi headers

### LINKED ISSUE #54: Windows: NVIDIA driver 610+ installs the Vulkan backend instead of CUDA, putting the AMD iGPU back in the auto device pool
## Summary

On Windows, machines running **NVIDIA driver 610 or newer** silently install the **Vulkan** llama.cpp backend instead of the CUDA one, even though `nvidia-smi` is present and healthy.

That matters beyond raw speed: the CUDA builds enumerate NVIDIA devices only, while the Vulkan build enumerates *every* Vulkan-capable device. On a system with an AMD APU, that puts the integrated GPU into `--list-devices` output and into the `auto` device-selection pool — the exact situation #35 / 0a10bf2 was meant to eliminate.

## Environment

- Windows, RTX 5070 Ti (Blackwell) + Ryzen 9 9900X (iGPU present)
- NVIDIA driver **610.47**, CUDA UMD 13.3
- atomic-agent 0.1.67 (fresh install)

## Symptom

Fresh install; `models` reports the 5070 Ti as the default device and model weights do load into its VRAM, but the discrete card sits at 0–1% utilisation during generation while the Ryzen iGPU shows activity.

Backend directory after install contains `ggml-vulkan.dll` and **no `ggml-cuda.dll`**:

```
ggml.dll
ggml-base.dll
ggml-cpu-cascadelake.dll
ggml-vulkan.dll          <-- Vulkan build
llama.dll
llama-common.dll
llama-server.exe
llama-server-impl.dll
mtmd.dll
```

Meanwhile `nvidia-smi` answers correctly, and fast (53 ms — well inside the 4 s probe timeout):

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 610.47                 KMD Version: 610.47        CUDA UMD Version: 13.3     |
+-----------------------------------------+------------------------+----------------------+
|   0  NVIDIA GeForce RTX 5070 Ti   WDDM  |   00000000:01:00.0  On |                  N/A |
```

## Root cause

NVIDIA driver 610 reworked the `nvidia-smi` header. `Driver Version` became `KMD Version`, and `CUDA Version` became **`CUDA UMD Version`**:

```
# driver 5xx
| NVIDIA-SMI 552.22   Driver Version: 552.22   CUDA Version: 12.6 |

# driver 610+
| NVIDIA-SMI 610.47   KMD Version: 610.47      CUDA UMD Version: 13.3 |
```

`parseDriverCudaVersion` matches the old field name literally (`src/local-llm/windows-backend-variant.ts:31`):

```ts
const match = output.match(/CUDA Version:\s*(\d+)\.(\d+)/i);
```

On a 610+ header that returns `null`, and `selectWindowsBackendAsset(null)` falls back to Vulkan — which is correct behaviour for "no NVIDIA driver detected", just fed a false premise. The CUDA gating added in 0a10bf2 is fine; it is simply never handed a version to gate on.

This is triggered by driver version, not by any particular machine, so every Windows box that has updated to a 610+ driver has been quietly downgraded to Vulkan since that driver shipped.

## Recovery on an already-affected machine

Fixing the parser is not enough on its own for an existing install — `isBackendDownloaded` only checks that `llama-server.exe` exists, and all four Windows zips ship that same filename. Delete the backend directory and run `atomic-agent models update` to force a clean re-download.

## Related

Adjacent gap, not fixed here: `checkForBackendUpdate` guards its variant-staleness check behind `current?.asset !== undefined` (`src/local-llm/backend-installer.ts:137`), so an install whose `backend-version.json` predates the `asset` field can never be detected as running the wrong compute backend while its tag matches the latest release.

sachin-detrax:
Closes #54.

NVIDIA driver 610 renamed the `nvidia-smi` header fields — `Driver Version` became `KMD Version`, and `CUDA Version` became `CUDA UMD Version`:

```
# driver 5xx
| NVIDIA-SMI 552.22   Driver Version: 552.22   CUDA Version: 12.6 |

# driver 610+
| NVIDIA-SMI 610.47   KMD Version: 610.47      CUDA UMD Version: 13.3 |
```

`parseDriverCudaVersion` matched `CUDA Version:` literally, so on any 610+ machine it returned `null` and `selectWindowsBackendAsset(null)` fell back to Vulkan by design.

The consequence is worse than losing CUDA throughput. The CUDA builds enumerate NVIDIA devices only; the Vulkan build enumerates everything, so an AMD APU lands back in `--list-devices` and in the `auto` selection pool — the condition 0a10bf2 removed. Reported on an RTX 5070 Ti + Ryzen 9900X box on driver 610.47, where the backend dir held `ggml-vulkan.dll` and no `ggml-cuda.dll` despite `nvidia-smi` answering in 53 ms with CUDA 13.3.

The gating logic from 0a10bf2 was correct; it was never handed a version to gate on.

## Change

One regex, allowing an optional qualifier between "CUDA" and "Version" so both header generations parse:

```diff
-  const match = output.match(/CUDA Version:\s*(\d+)\.(\d+)/i);
+  const match = output.match(/CUDA(?:\s+\w+)?\s+Version:\s*(\d+)\.(\d+)/i);
```

Deliberately not loosened to bare `CUDA`, so an unrelated line carrying a version number can't be mistaken for the driver's CUDA runtime ceiling.

`nvidia-smi-vram.ts` is unaffected — it uses `--query-gpu` machine-readable output, not the header.

## Verification

- New test covers the 610.47 header; it fails on the old regex and passes on the new one.
- Existing coverage for the 5xx header and the `cuda version:13.3` spacing case still passes.
- `src/local-llm` suite: 116/116 pass. `tsc --noEmit` clean.
- Confirmed on the reporting machine: after this change plus a forced backend re-download, `ggml-cuda.dll` is installed and generation runs on the 5070 Ti.

## Note for already-affected installs

The parser fix alone does not repair a machine that already has the Vulkan backend on disk — `isBackendDownloaded` only checks that `llama-server.exe` exists, and every Windows zip ships that same filename. Deleting the backend directory before `models update` forces a clean re-download.

Adjacent gap left alone here: `checkForBackendUpdate` guards variant staleness behind `current?.asset !== undefined` (`src/local-llm/backend-installer.ts:137`), so an install whose `backend-version.json` predates the `asset` field can never be flagged as running the wrong compute backend while its tag matches latest. Happy to fix in a follow-up if you want it in the same release.

### DIFF
--- a/src/local-llm/windows-backend-variant.test.ts
+++ b/src/local-llm/windows-backend-variant.test.ts
@@ -33,6 +33,18 @@ describe("parseDriverCudaVersion", () => {
     });
   });
 
+  it("reads the `CUDA UMD Version` field used by driver 610+", () => {
+    // Driver 610.x reworked the header: `Driver Version` became `KMD
+    // Version` and `CUDA Version` became `CUDA UMD Version`. Matching
+    // `CUDA Version:` literally returned null on those boxes, which sent
+    // an RTX 5070 Ti to the Vulkan build.
+    expect(
+      parseDriverCudaVersion(
+        "| NVIDIA-SMI 610.47                 KMD Version: 610.47        CUDA UMD Version: 13.3     |",
+      ),
+    ).toEqual({ major: 13, minor: 3 });
+  });
+
   it("returns null when the field is absent", () => {
     expect(parseDriverCudaVersion("no gpu here")).toBeNull();
   });
--- a/src/local-llm/windows-backend-variant.ts
+++ b/src/local-llm/windows-backend-variant.ts
@@ -21,14 +21,23 @@ export interface CudaVersion {
 }
 
 /**
- * Parse the `CUDA Version: X.Y` field from `nvidia-smi` header output.
- * This value is the **maximum** CUDA runtime version the installed
- * driver supports (not the CUDA toolkit version), which is exactly what
- * we need to decide which prebuilt CUDA backend will load. Returns null
- * when the field is absent.
+ * Parse the CUDA version field from `nvidia-smi` header output. This
+ * value is the **maximum** CUDA runtime version the installed driver
+ * supports (not the CUDA toolkit version), which is exactly what we need
+ * to decide which prebuilt CUDA backend will load. Returns null when the
+ * field is absent.
+ *
+ * The field name is not stable across drivers. Up to 5xx/6xx-early the
+ * header reads `Driver Version: X  CUDA Version: Y`; driver 610 renamed
+ * the pair to `KMD Version: X  CUDA UMD Version: Y`. Matching
+ * `CUDA Version:` literally returned null on 610 boxes, so every machine
+ * on a current driver silently fell back to the Vulkan build — which is
+ * the only configuration that enumerates an AMD APU alongside the
+ * NVIDIA card. The optional word tolerates that rename (and any future
+ * qualifier) without loosening the match to bare "CUDA".
  */
 export function parseDriverCudaVersion(output: string): CudaVersion | null {
-  const match = output.match(/CUDA Version:\s*(\d+)\.(\d+)/i);
+  const match = output.match(/CUDA(?:\s+\w+)?\s+Version:\s*(\d+)\.(\d+)/i);
   if (!match) return null;
   const major = Number(match[1]);
   const minor = Number(match[2]);


Ooooze:
Thanks for the thorough writeup and repro — verified the regex change against the old header, the new `CUDA UMD Version` header, and the existing case-insensitive/spacing test, all backtrack correctly to the right major.minor. Merging.

PR #64, opened 2026-08-06
feat(tui): reopenable cloud model picker, wired to /model

### LINKED ISSUE #62: Cloud model picker is not reachable after initial provider setup
**Current behavior**

When a user configures an OpenAI-compatible cloud provider (custom base URL, e.g. vLLM or LM Studio), the setup wizard fetches `{baseUrl}/v1/models` and shows a model list to pick from (#41). After the wizard completes, that list is gone for good:

- The Cloud pane of the LLM panel renders exactly one chat model row for an `openai-compatible` provider: the one already selected (`chatModelsForProvider` in `src/tui/llm-panel/llm-panel-row-builders.ts`). Pressing Enter on it re-selects the same model.
- `/model` only jumps to the LLM tab; it takes no argument and opens no picker (`src/tui/commands/slash-command-handler.ts`).
- The only in-TUI way to change the model is to re-run the provider wizard, which restarts at the API key screen and does not prefill the base URL. Pressing Enter through the empty base URL step silently resets it to `https://api.openai.com`, clobbering the custom endpoint.

Effectively the user is locked into the model chosen during first-time setup unless they re-run the whole wizard or edit config.json by hand.

**Expected behavior**

A model picker that can be reopened at any time, which is the common pattern in agent CLIs: `/model` and Enter on the current model row should open the provider's model list, with the current model marked, and switching should take effect mid-session without re-entering the API key or the base URL.

**Implementation note**

The building blocks already exist: `fetchOpenAiCompatModels` / its 1h cache (added in #41) and the wizard's pick UI. The main work is exposing that picker outside the wizard: trigger it from the `cloudChatModel` row and from `/model`, and route the selection through the existing `onProvidersSelectChatModel` callback, which already persists and applies the change.

**Related**

While here, the configure-mode wizard should prefill the stored base URL instead of defaulting to `https://api.openai.com` on an empty field; today a quick re-run can silently break a custom endpoint.

sosidudku1:
Closes #62. Based on the branch of the base-URL prefill fix; merge that one first.

## Problem

After first-time setup the `/v1/models` list is unreachable: the Cloud pane renders a single model row for an openai-compatible provider and Enter re-selects it, `/model` only jumps to the LLM tab, and the only in-TUI way to switch models is re-running the wizard from the API key screen.

## Change

- **New `modelPicker` modal in the LLM panel.** Enter on the model row of an openai-compatible provider reopens the server's model list: 12-row window with wrap-around cursor, the current model marked, count in the footer. Enter switches the model mid-session through the existing `providers_select_chat_model` path (which already persists and applies the change), Esc cancels. An unreachable or unauthorized endpoint degrades to a closable error box instead of trapping the user.
- **`/model` now opens the picker** for the active text provider by emitting `providers_chat_model_picker_requested`. The request no-ops for curated kinds (their models are already first-class rows), so the previous tab-jump behavior is unchanged there.
- **The async fetch is owned by `ProvidersOrchestrator`**, following the existing bus-subscription pattern, and reuses `fetchOpenAiCompatModels` with its 1h per-URL-and-key cache from #41. Stale responses arriving after the picker was closed or reopened for another provider are dropped in the reducer.

Out of scope, possible follow-ups: fuzzy search inside the picker, recent-models ordering, a global hotkey.

## Testing

10 new tests: reducer transitions (open/loaded with cursor on the current model/stale-response guards/failed) and modal keys (wrap-around navigation, select-and-close routing, Esc, loading swallows keys, error closes on Enter). The pre-named `/model` slash test now asserts the picker request it was named after. `src/tui` plus `src/llm/provider`: 730 passing, the 6 remaining failures are the pre-existing set documented in #50, verified identical on clean main. `tsc --noEmit` clean.


src/tui/llm-panel/llm-panel-state.ts
```diff
@@ -27,6 +42,7 @@ export interface LlmPanelState {
    * keyboard.
    */
   externalUrlDraft: string | null;
+  modelPicker: LlmModelPickerState | null;
```
Ooooze: This is the new picker state. Per the top-level comment: consider whether this belongs on `providersPanel` instead (see `wizard`/`removeConfirm` precedent), and either way it needs to be added to `llmTabBusy` in `app-key-bindings.ts` so Tab/Ctrl+B don't leak past the open modal.

src/tui/providers/providers-orchestrator.ts
```diff
@@ -72,6 +76,39 @@ export class ProvidersOrchestrator {
...
+  async openChatModelPicker(providerId: string | null): Promise<void> {
+    const config = getConfig();
+    const resolved = resolveLlmConfig(config);
+    const id = providerId ?? resolved.activeTextProvider;
+    if (!id) return;
+    const provider = resolved.providers.find((p) => p.id === id);
+    const fileEntry = config.llm?.providers.find((e) => e.id === id);
+    if (!provider || provider.kind !== "openai-compatible") return;
+    const baseUrl = fileEntry?.baseUrl ?? OPENAI_COMPAT_DEFAULT_BASE_URL;
+    this.bus.emit({
```
Ooooze, line 101: `ProvidersOrchestrator` emitting `llm_model_picker_*` (a type owned by `llm-panel-actions.ts`) is the only place in the whole TUI where an orchestrator reaches into another panel's action namespace — every other orchestrator (tasks/skills/mcp/memory/telegram/local-models) only emits its own namespace or the shared global actions (`runtime_info`, `ui_mode_set`). Given `LlmPanelModals` already renders `providersPanel.wizard`/`removeConfirm` directly, moving this picker's state into `providersPanel` and renaming these to `providers_chat_model_picker_*` would be more consistent with the existing pattern in this same file.

src/tui/llm-panel/llm-panel-reducer.ts
```diff
@@ -70,6 +70,71 @@ export function reduceLlmPanelAction(
...
+            models: [],
+            cursor: 0,
+            error: null,
+          },
+        },
+      };
+    case "llm_model_picker_loaded": {
+      // A stale fetch (picker closed or reopened for another provider)
+      // must not resurrect or repoint the modal.
+      if (!panel.modelPicker || panel.modelPicker.providerId !== action.providerId) {
```
Ooooze: Stale-response guard is keyed only on `providerId`. Closing the picker (Esc) and reopening it for the *same* provider before the first fetch resolves means this check can't distinguish the two in-flight requests — the old one's result will still be accepted here. Low-impact today thanks to the 1h cache in `fetchOpenAiCompatModels`, but a generation/request-id would make the guard actually correct rather than incidentally correct.

Ooooze (changes requested):
Reviewed locally on `feat/reopenable-model-picker` (worktree checkout, `tsc --noEmit` clean, targeted vitest run green aside from a pre-existing flaky catalog-pricing test in `llm-panel-selectors.test.ts` that is unrelated to this change).

## Blocking: modal doesn't participate in the tab-busy gate → Tab/Ctrl+B leak out of it

`src/tui/app-key-bindings.ts` computes `llmTabBusy` from a fixed list of "a modal owns the keyboard" flags:

```ts
const llmTabBusy =
    state.uiMode === "debug" &&
    state.activeTab === "llm" &&
    (state.providersPanel.wizard !== null ||
      state.providersPanel.removeConfirm !== null ||
      state.localModelsPanel.mode === "backendUpdate" ||
      state.localModelsPanel.pull !== null ||
      state.localModelsPanel.removeConfirmId !== null ||
      state.localModelsPanel.embeddingRemoveConfirmId !== null ||
      state.localModelsPanel.embeddingOnboardingPrompt !== null ||
      state.llmPanel.stopLocalDaemonsPrompt !== null);
```

`state.llmPanel.modelPicker` is missing from this list. `debugTabBusy` (which folds `llmTabBusy` in) gates both the Tab/Shift+Tab nav-cycle and the Ctrl+B nav-cycle inside `handleAppKey`, and `handleAppKey` runs *before* `handleLlmPanelKey` / `handleLlmModalKey` in `tui-app.tsx`'s `useInput` dispatch:

```ts
const appHandled = handleAppKey(input, key, {...});
if (appHandled) return;           // <-- short-circuits before handleLlmPanelKey
...
if (llmTabActive) { handleLlmPanelKey(...); return; }
```

Concretely: while the model picker is open, pressing **Tab** or **Ctrl+B** never reaches `handleLlmModalKey` at all — it's swallowed by the global nav-cycler, which contradicts the modal's own "the modal owns the keyboard" comment in `llm-panel-modal-key-bindings.ts` and the exact pattern this codebase already uses for the MCP tab's modals (see the `mcpTabBusy` doc-comment a few lines above `llmTabBusy` in the same file — "the panel must keep capturing letter/Tab keys"). None of the new tests exercise this path since `llm-panel-model-picker.test.ts` calls `handleLlmModalKey` directly, bypassing `handleAppKey`.

Fix: add `state.llmPanel.modelPicker !== null` to `llmTabBusy` (and, while touching this list, `state.llmPanel.externalUrlDraft !== null` is missing too — pre-existing gap, but worth closing in the same pass since this PR is already extending the busy-gate surface).

## Architecture nit worth addressing before merge

`LlmModelPickerState` lives on `llmPanel` (`llm-panel-state.ts`) and is driven entirely by `ProvidersOrchestrator`, which emits `llm_model_picker_opened/loaded/failed` — action types owned by a different module (`llm-panel-actions.ts`) than the one dispatching them.

That's inconsistent with the two closest siblings in the very file this PR touches: `LlmPanelModals` already renders `state.providersPanel.wizard` and `state.providersPanel.removeConfirm` directly — both are provider-shaped modal state that happens to render in the LLM tab, driven by `providers_*`-namespaced actions from the same orchestrator. A grep across every other TUI orchestrator (`tasks`, `skills`, `mcp`, `memory`, `telegram`, `local-models`) confirms none of them emit an action type owned by a different panel's slice; they only emit their own namespace plus the handful of genuinely global actions (`runtime_info`, `ui_mode_set`, `system_message`).

Suggest moving `LlmModelPickerState` into `providers-panel-state.ts` / `providers-actions.ts` / `providers-reducer.ts` (renaming the action types to `providers_chat_model_picker_*`) and having `LlmPanelModals` read `state.providersPanel.modelPicker`, matching the existing `wizard`/`removeConfirm` pattern. This isn't just cosmetic — `providersPanel.wizard`/`removeConfirm` are already co-located in the `llmTabBusy` list precisely because they live in the same slice; `modelPicker` landing in a different slice is plausibly *why* it was left out of that list in the first place.

## Minor / non-blocking

- `llm-panel-reducer.ts`'s stale-response guard on `llm_model_picker_loaded` keys only on `providerId`, not a request/generation id. Close the picker (Esc) while a fetch is in flight, then immediately reopen it for the *same* provider before the first fetch settles, and the reducer can't tell the two fetches apart (`panel.modelPicker.providerId === action.providerId` will be true for both). Low real-world impact today since `fetchOpenAiCompatModels` caches by URL+key for 1h so both responses are identical, but the guard itself doesn't actually protect against this ordering — flagging in case the cache TTL or key ever changes.
- `ProvidersOrchestrator.openChatModelPicker` silently no-ops when `/model` is invoked while the active text provider isn't `openai-compatible` (local-llama / openrouter / aimlapi). Matches the PR description ("no-ops for curated kinds"), but a `providers_status` line would make it discoverable instead of just landing on the LLM tab with nothing else happening.

Happy to re-review once the busy-gate fix lands — that's the only thing I'd consider a hard blocker here.

sosidudku1:
All three addressed, thanks for the review.

1. **Namespace/ownership.** The picker state moved to `providersPanel` (`chatModelPicker`) and the actions are now `providers_chat_model_picker_*`, so `ProvidersOrchestrator` only emits its own namespace. The reducer cases moved to `providers-reducer.ts` alongside `wizard`/`removeConfirm`, and the tests moved with them to `src/tui/providers/providers-chat-model-picker.test.ts`.

2. **`llmTabBusy`.** Added the open picker. While there I also added `llmPanel.externalUrlDraft`, which has been missing from that guard since #50, so Tab/Ctrl+B leaked past the external URL editor too.

3. **Stale-response guard.** Now keyed on a generation counter bumped on every open (`chatModelPickerGeneration` on the orchestrator, mirrored into the picker state) rather than on `providerId`, so the close-and-reopen-same-provider race you described is rejected properly rather than incidentally. There is a test for exactly that sequence.

`tsc` clean; `src/tui` plus `src/llm/provider` show the same pre-existing failures as main (the two `LlmHealthPoller` ones are timing-flaky and pass in isolation).

Ooooze (approved):
Follow-up commit `1f80520` verified locally (checked out worktree, `tsc --noEmit` clean, targeted vitest run green apart from the pre-existing unrelated flaky pricing test in `llm-panel-selectors.test.ts`):

- `llmTabBusy` now covers `providersPanel.chatModelPicker` and `llmPanel.externalUrlDraft` — Tab/Ctrl+B no longer leak past the open modal.
- Picker state moved to `providersPanel` / `providers_chat_model_picker_*`, matching the existing `wizard`/`removeConfirm` precedent; `providers-reducer.ts` owns it end to end.
- Stale-response guard is now keyed on a monotonic `generation` counter (better than the request-id I suggested), with a direct regression test for the same-provider reopen race.

All three points addressed. Approving.

Ooooze (approved):
Follow-up commit `1f80520` verified locally (checked out worktree, `tsc --noEmit` clean, targeted vitest run green apart from the pre-existing unrelated flaky pricing test in `llm-panel-selectors.test.ts`):

- `llmTabBusy` now covers `providersPanel.chatModelPicker` and `llmPanel.externalUrlDraft` — Tab/Ctrl+B no longer leak past the open modal.
- Picker state moved to `providersPanel` / `providers_chat_model_picker_*`, matching the existing `wizard`/`removeConfirm` precedent; `providers-reducer.ts` owns it end to end.
- Stale-response guard is now keyed on a monotonic `generation` counter (better than the request-id I suggested), with a direct regression test for the same-provider reopen race.

All three points addressed. Approving.

PR #67, opened 2026-08-07
fix(providers): serve the full model catalog and stop reading API silence as no

sosidudku1:
Reported internally: the cloud model picker offers only a handful of models. Against the live APIs today the picker shows **12 of 305** usable OpenRouter models and **13 of 337** aimlapi models.

Three independent causes.

## 1. The live catalogs were capped

`MAX_PICKS` cut the list to 12 (OpenRouter) and 32 (aimlapi) after the curated ids. That cap made sense for an unfilterable list; it silently hides most of the catalog. Both caps are gone. Curated ids keep their hand-picked order, the OpenRouter tail keeps its score order, and the aimlapi tail is now sorted alphabetically instead of following map insertion order.

## 2. aimlapi changed its response shape, and the parser matched nothing

Two breaking changes on their side, neither of which failed loudly:

- rows are now typed `openai/chat-completions` instead of `chat-completion`, so the equality check matched zero rows;
- the `features` array that advertised tool support was removed entirely, so the "can this model call tools" check rejected everything that survived.

The net effect was a live refresh that always returned nothing and fell back to the 13-entry offline list, which also explains the `price unknown` labels in the picker.

Type matching is now suffix-based (`*/chat-completions`, plus the older spellings), which survives another vendor-prefix rename while still excluding the video, image, `anthropic/messages`, and `responses/submit` families that share the same payload.

## 3. Absent was read as negative

Both fetchers now distinguish "the API says this model has no tools" from "the API says nothing". Only the former drops a model. This is the generic form of the aimlapi breakage: a provider dropping a capability field can no longer empty our catalog. OpenRouter has the same guard even though its `supported_parameters` is present today.

## Review follow-up

The second commit addresses review findings. The 12-row viewport around the cursor now lives in `renderPickList` itself, so the OpenRouter and aimlapi picker phases window their rows instead of painting the full 300+ row catalog into the terminal, with a position counter once the list outgrows the window. Both fetchers skip null and scalar rows in `data`, so one bad row can no longer throw and drag the whole live catalog into the static fallback. The fallback tests now reset the module-level cache (`vi.resetModules` plus a dynamic import) and assert the exact offline list, so they test the fallback rather than a leftover cache. The stale JSDoc describing the removed type/features filter is rewritten, and a comment explains why vision silence deliberately still reads as "no vision".

## Verification

Ran the parsers against the live payloads captured from both providers: **337 aimlapi picks** (was 13) and **305 OpenRouter picks** (was 12).

Tests: 16 new across both commits. Cap removal on both providers, offline fallback on network error and on a malformed payload (against a fresh module cache), null rows skipped, the current aimlapi shape verbatim (including the non-chat families that must stay out), explicit no-tools still dropped, silent-API models kept, rows without a usable id ignored, alphabetical tail ordering, and 12-row windowing of 300+ model lists in every wizard picker phase. `tsc` clean. Full suite shows the same pre-existing failures as main.

## Scope note

Full catalog does not mean every id the providers return: `scoreChat` keeps the deliberate exclusions of `anthropic/*` and `/gemini/i`, and the catalog tests pin both. The recorded reason is the tool transport, not model preference. The cloud path drives models through the OpenAI chat-completions wire shape with native tools, and the adapter seam explicitly marks Anthropic and Gemini adapters as future work (`src/llm/provider/adapters/tool-call-adapter.ts`, since 5c7c05e). Claude additionally sits on a different surface at aimlapi: its rows are typed `anthropic/messages` and 404 on `/v1/chat/completions`. Gemini started out score-boosted and was removed in c0a73b4 together with the recovery handling for cloud models that emit tool calls as plain content, which points at tool-call reliability through the OpenAI-shape adapter. So this PR ships the full set of models the runtime can actually drive today: full minus Claude and Gemini. When dedicated adapters land, lifting the two filters should be part of that change.

## Note

Until the picker gains text filtering, these lists are long. That work is the companion branch; merging this one first is fine, since the wizard picker now windows its rows too.

### DIFF
--- a/src/llm/provider/aimlapi/fetch-aimlapi-chat-catalog.ts
+++ b/src/llm/provider/aimlapi/fetch-aimlapi-chat-catalog.ts
@@ -14,10 +14,14 @@ type AimlapiApiModel = {
    * Capability flags. Present in the pre-2026-08 response shape; the
    * current API dropped the field entirely, so every reader must treat
    * it as optional rather than assuming its absence means "no support".
+   *
+   * The current shape's `tags` array is deliberately NOT modelled here:
+   * as of 2026-08 it only carries playground grouping and pricing tier
+   * labels (`playground:chat`, `tier:tier_2`), never capability
+   * markers (the gpt-4o family shows up without any vision tag), so
+   * nothing can be derived from it.
    */
   features?: readonly string[];
-  /** Current shape: `playground:chat`, `tier:tier_2`, and similar. */
-  tags?: readonly string[];
   info?: {
     contextLength?: number;
     context_length?: number;
@@ -89,10 +93,24 @@ function readToolSupport(m: AimlapiApiModel): "none" | "basic" | "parallel" {
   return readAdvertisedToolSupport(m) ?? "basic";
 }
 
+/**
+ * Unlike tools, silence deliberately reads as "no vision", and the cost
+ * of that is real, not cosmetic. `supportsVision` becomes the
+ * capability bit on the synthesised {@link ModelCatalogEntry}, which is
+ * what the provider layer reads to gate image input
+ * (`ProviderCapabilities.vision`, whose `false` makes `vision.describe`
+ * raise `VisionUnsupportedError`), so a live-only id whose row stays
+ * silent loses `vision.describe` entirely, not just a "vision" badge in
+ * the picker. We still prefer underclaiming: advertising vision on a
+ * text-only route breaks at request time with a rejected image payload,
+ * while a conservative entry keeps the model usable for text and, unlike
+ * the tools case, can never empty the catalog. Curated ids are
+ * unaffected (their hand-verified static entry wins in
+ * `entryFromLiveModel`), and the current response offers no better
+ * signal to read: `features` is gone, `tags` and `info` carry no
+ * capability fields (see {@link AimlapiApiModel}).
+ */
 function readVisionSupport(m: AimlapiApiModel): boolean {
-  // Unlike tools, silence deliberately reads as "no vision": overclaiming
-  // would invite image input that the backend rejects, while underclaiming
-  // only hides a badge and never empties the catalog.
   return (m.features ?? []).some((f) => f.includes(".vision"));
 }
 
@@ -127,13 +145,20 @@ function formatCtx(tokens: number): string {
   return `${tokens}`;
 }
 
+let staticPicks: readonly AimlapiChatPick[] | null = null;
+
 function picksFromStaticCatalog(): readonly AimlapiChatPick[] {
+  // Memoized: downstream lookup caches key themselves on the array
+  // reference (see providers-model-options), so the offline fallback must
+  // return a stable array rather than a fresh one per call.
+  if (staticPicks) return staticPicks;
   const out: AimlapiChatPick[] = [];
   for (const id of AIMLAPI_CHAT_MODEL_ORDER) {
     const entry = AIMLAPI_MODELS_CATALOG.get(id);
     if (!entry || entry.kind !== "chat") continue;
     out.push({ id, label: labelForPick(id, entry), entry });
   }
+  staticPicks = out;
   return out;
 }
 
--- a/src/llm/provider/aimlapi/fetch-aimlapi-chat-catalog.ts
+++ b/src/llm/provider/aimlapi/fetch-aimlapi-chat-catalog.ts
@@ -63,7 +63,7 @@ function readContextLength(m: AimlapiApiModel): number {
  * `undefined` means "the response says nothing about it", which is the
  * case for the whole current catalog: the `features` array was removed
  * and nothing replaced it. Callers must not read that silence as "no
- * tools" — doing so filtered out all 337 chat models and silently pinned
+ * tools": doing so filtered out all 337 chat models and silently pinned
  * the picker to the offline list.
  */
 function readAdvertisedToolSupport(
@@ -90,6 +90,9 @@ function readToolSupport(m: AimlapiApiModel): "none" | "basic" | "parallel" {
 }
 
 function readVisionSupport(m: AimlapiApiModel): boolean {
+  // Unlike tools, silence deliberately reads as "no vision": overclaiming
+  // would invite image input that the backend rejects, while underclaiming
+  // only hides a badge and never empties the catalog.
   return (m.features ?? []).some((f) => f.includes(".vision"));
 }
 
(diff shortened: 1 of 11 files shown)


src/tui/components/providers-wizard.tsx
```diff
@@ -43,12 +43,26 @@ function maskedKey(buffer: string): string {
...
+  // OpenRouter/aimlapi catalogs run past 300 rows, and an unwindowed map
+  // would paint them all into the terminal at once.
+  const total = options.length;
+  const clamped = Math.min(Math.max(cursor, 0), Math.max(0, total - 1));
+  const start = Math.min(
+    Math.max(0, clamped - Math.floor(PICK_WINDOW / 2)),
+    Math.max(0, total - PICK_WINDOW),
+  );
+  const visible = options.slice(start, start + PICK_WINDOW);
+  const position = total > PICK_WINDOW ? ` (${clamped + 1}/${total})` : "";
```
Ooooze: Windowing here fixes the rendering cost but not the build cost. Both cloud phases call `listOpenRouterChatModels()` / `listAimlapiChatModels()` in the component body, which formats a label for all 337 rows before we slice 12 of them — and each label goes through `formatAimlapiChatModelDetails`, which does a linear `listAimlapiChatPicks().find(...)`. That is ~113k comparisons per frame (it was 12x12 before the cap came off), re-run on every j/k keystroke.

Cheapest fix is in `providers-model-options.ts`: build a `Map<id, entry>` once instead of `.find` per row in `resolveOpenRouterCatalogEntry` / `resolveAimlapiCatalogEntry`. Alternatively pass the window offset down so the label formatting only runs for visible rows.

Also: with these additions the file is 312 lines, past the 300-line ceiling in AGENTS.md. `renderPickList` + `PICK_WINDOW` look like the natural extraction into their own module.

src/tui/components/providers-wizard.tsx
```diff
@@ -150,16 +168,11 @@ function CompatChatModelStep(props: {
...
-      Math.max(0, cursor - Math.floor(PICK_WINDOW / 2)),
-      Math.max(0, picks.length - PICK_WINDOW),
-    );
     return renderPickList(
       `Chat model — ${picks.length} from ${baseUrl}/v1/models`,
-      picks.slice(start, start + PICK_WINDOW).map((id) => ({ label: id })),
-      cursor - start,
-      `↑/↓ move (${cursor + 1}/${picks.length}) · Enter select · type to enter an id by hand · Esc cancel`,
+      picks.map((id) => ({ label: id })),
+      w.cursor,
```
Ooooze: Two small regressions from moving the window into `renderPickList`:

1. The position counter is now conditional on `total > PICK_WINDOW`, so this step loses the `(5/30)` readout it always showed for short lists, and for long ones the counter moved from the head of the hint to its tail.

2. This step used to clamp with `Math.min(w.cursor, picks.length - 1)` before rendering. Clamping now happens only inside `renderPickList`, while `advanceWizardPhase` still selects via `models[wizard.cursor] ?? models[0]`. If the cursor ever outruns the list (catalog shrinking under a refresh between renders), the last row is highlighted but the first one is picked. Worth clamping the cursor in state rather than at paint time.

Ooooze (commented):
Nice diagnosis — three genuinely independent causes, and the fixes are structural rather than patches. Suffix matching on `type`, keeping `undefined` apart from `"none"`, skipping null rows, and `vi.resetModules()` in the fallback tests (which previously asserted against a leftover cache and proved nothing) are all the right calls.

Two things I'd resolve before this lands, both consequences of removing the cap rather than problems with the parsing work itself.

**Model option lists are O(n^2) per render.**
`src/tui/providers/providers-model-options.ts` is not in this diff, but it is where the cap removal bites: `formatOpenRouterChatModelDetails` / `formatAimlapiChatModelDetails` resolve their entry with a linear `listXChatPicks().find(...)`, once per row, for a list that is now 305-337 rows instead of 12. That is ~113k comparisons per frame, re-run on every j/k keystroke, and the window in `renderPickList` does not help because it slices after every label is built. Swapping the `.find` for a `Map<id, entry>` lookup is a one-line change and keeps the picker responsive.

**Reaching a model 250 rows down.**
`providers-wizard-key-bindings.ts` only advances one row at a time (`(cursor + 1) % len`) — no PageUp, no Home, no jump. Windowing makes a 305-row catalog renderable, not navigable. Either land this together with #68 (typed filtering) or add a page-jump here; on its own this is arguably worse to operate than the curated 12.

**On the description rather than the code:** `scoreChat` still returns -1 for `anthropic/*` and `/gemini/i`, and the existing test pins that. So "the full model catalog" is the full catalog minus Claude and Gemini — which is what most people will look for first. Worth calling that exclusion out explicitly (and why it exists) so it doesn't come back as a follow-up issue.

src/llm/provider/aimlapi/fetch-aimlapi-chat-catalog.ts
```diff
@@ -32,18 +57,42 @@ function readContextLength(m: AimlapiApiModel): number {
...
+ * `/v1/chat/completions`, which is the tool-calling surface, and a model
+ * that turns out not to support tools fails loudly at request time
+ * rather than being invisible in the picker.
+ */
+function readToolSupport(m: AimlapiApiModel): "none" | "basic" | "parallel" {
+  return readAdvertisedToolSupport(m) ?? "basic";
+}
+
 function readVisionSupport(m: AimlapiApiModel): boolean {
+  // Unlike tools, silence deliberately reads as "no vision": overclaiming
```
Ooooze: The comment says underclaiming "only hides a badge", but `supportsVision` flows into `ModelCatalogEntry` -> `ProviderCapabilities`, and `registerVisionTools` / `LlamaServerProvider.describeImage` gate on it. So for any live id not covered by the static catalog, vision silence does not hide a badge — it disables `vision.describe` for that model entirely. Since `features` is gone from the current response, that is now every new live id.

Curated ids are safe (`entryFromLiveModel` returns the static entry), so this is not a regression against today's picker, but the reasoning in the comment undersells the consequence.

Related: this PR adds `tags?: readonly string[]` to `AimlapiApiModel` and never reads it. Either derive the capability signal from `tags` (it carries `playground:chat` and friends) or drop the field.

sosidudku1:
Thanks for the thorough pass. All addressed in the new commit:

**A)** Entry resolution is an id-to-entry Map now, keyed on the picks array reference and rebuilt only when a refresh swaps the array (the offline fallback is memoized so its reference is stable too). On top of that, option labels are lazy getters, so only the 12 rows the viewport paints get formatted; key handling reads ids and lengths only. Same effect as passing an offset, without the signature churn.

**B)** `renderPickList` + `PICK_WINDOW` moved to `src/tui/components/wizard-pick-list.tsx`. While splitting I found `providers-wizard-key-bindings.ts` was over the cap too (335), so the phase helpers moved to `providers-wizard-phases.ts`; every touched file is back under 300.

**C)** Both regressions fixed. The counter is unconditional again and sits right after the movement keys (`j/k move (5/30) · ...`), pinned by a test on the 3-row kind list. The cursor clamps in state via a shared `clampCursor` used by movement, page jumps and selection; a test shrinks the catalog under cursor 25 and asserts Enter picks the highlighted last row. The wizard has no list-changed event (the cache swaps in the background), so clamping at every key event is the reducer-level equivalent.

**D)** PgUp/PgDn jump one `PICK_WINDOW` and Home/End hit the edges, both in the cloud pick phases and in the openai-compatible discovered list, with tests for the jumps, the tail clamp, and `j` staying a printable character in the compat list. Typed filtering for the wizard pickers is the next PR in this series.

**E)** Comment rewritten to say what silence actually costs: `supportsVision` is the capability bit behind `ProviderCapabilities.vision`, which turns `vision.describe` into `VisionUnsupportedError` for live-only ids, not a hidden badge. On `tags`: I pulled the live `/v1/models` payload; across all 337 chat rows tags carry only `playground:*` grouping and `tier:*` pricing labels, and known-vision rows (the gpt-4o family) have no vision marker, so there is no capability signal to derive. The field was unread, so it is removed, with the finding documented next to `features`.

**F)** Agreed, the PR description now carries a scope note (added above in the description).

Unrelated heads-up: `llm-panel-selectors.test.ts` fails on main too (static qwen3.7-max pricing is 1.25/3.75, the test expects $2.5/$7.5); same before and after this branch.

PR #68, opened 2026-08-07
feat(tui): filter the model picker by typing, and stop it leaving stray glyphs

sosidudku1:
Companion to #67, which raises the catalogs to 337 and 305 models. Based on that branch; merge #67 first and this retargets to main.

## Filtering

Typing in the picker filters the list by substring, case-insensitively. Backspace trims the query, arrows and Enter operate on the filtered rows, and the cursor resets to the top on every keystroke so it can never point at a row that was just filtered away. Backspace with an empty query is a no-op, so it cannot reset the cursor. Ctrl and Meta combos are not typed into the filter; the modal swallows them like every other key, and Ctrl+C keeps working only because `handleAppKey` runs before the modal handler ever sees the key.

The counter reads `3/12 of 337` when a filter is active, and `no match` on an empty result set.

## Two repaint bugs

Both were visible in a screenshot of the panel before this change: model rows rendering as `toolss` and `toolsls`, and the header reading `Press ←/→ to switch modeternal llama.cpp`.

1. **The box changed height between renders.** Ink repaints a frame by overwriting the previous one line for line, so a shorter frame leaves the tail of the taller one on screen. The window is now fixed height and pads with blank rows when the result set is shorter.

2. **Rows were keyed by model id.** Ids repeat across a re-filter, and duplicate keys made Ink reuse the previous row's text, which left half-erased names behind (reported as "the last letter of the model sticks"). Rows are now keyed by slot.

The window-start computation is also clamped defensively for lists shorter than the window. The previous expression already produced correct slices in that case; the clamp just makes the invariant explicit instead of implicit.

## Testing

12 new tests: query narrows the rows and resets the cursor, case-insensitive substring matching, empty query shows everything, no-match yields an empty list, printable keys dispatch a query update, backspace trims, backspace on an empty query is a no-op that keeps the cursor put, arrows wrap within the filtered list, Enter selects from the filtered list, Enter on an empty result does nothing, Esc closes with a query active, Ctrl combos are not typed.

`src/tui/providers`: 39 passing. `tsc` clean. Full suite shows the same pre-existing failures as main.

### DIFF
--- a/src/tui/components/llm-panel-modals.tsx
+++ b/src/tui/components/llm-panel-modals.tsx
@@ -6,7 +6,43 @@ import { ProvidersWizard } from "./providers-wizard.js";
 import { parseExternalUrl } from "../llm-panel/llm-panel-modal-key-bindings.js";
 import { filteredPickerModels } from "../providers/providers-panel-state.js";
 
-export function LlmPanelModals({ state }: { state: TuiState }): ReactElement | null {
+/** Upper bound for the picker's list window (roomy terminals). */
+const PICKER_MAX_WINDOW = 12;
+/** Never shrink the list window below this many rows. */
+const PICKER_MIN_WINDOW = 3;
+/**
+ * Picker box rows that are not list rows: two border lines, the title,
+ * the filter/status line, the footer, and the box's bottom margin.
+ */
+const PICKER_CHROME_ROWS = 6;
+
+/**
+ * List rows the picker shows. Sized from the tab's row budget so short
+ * terminals get a smaller (but still fixed) window instead of a frame
+ * that outgrows the screen.
+ */
+function pickerWindowRows(maxRows: number | undefined): number {
+  if (maxRows === undefined) return PICKER_MAX_WINDOW;
+  return Math.max(
+    PICKER_MIN_WINDOW,
+    Math.min(PICKER_MAX_WINDOW, maxRows - PICKER_CHROME_ROWS),
+  );
+}
+
+/** `count` blank lines that hold a box at its fixed height. */
+function blankRows(count: number): ReactElement[] {
+  return Array.from({ length: count }, (_unused, i) => (
+    <Text key={`pad-${i}`}> </Text>
+  ));
+}
+
+export function LlmPanelModals({
+  state,
+  maxRows,
+}: {
+  state: TuiState;
+  maxRows?: number;
+}): ReactElement | null {
   if (state.providersPanel.wizard) {
     return <ProvidersWizard wizard={state.providersPanel.wizard} />;
   }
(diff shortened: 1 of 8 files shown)


Ooooze (commented):
Read through the diff against the current tree. The state design is the right shape: `query` on the picker plus a derived `filteredPickerModels` selector means the filter can never drift out of sync with `models`, and resetting the cursor on every keystroke removes the whole class of "cursor points at a row that was just filtered away" bugs. The key routing also holds up — `chatModelPicker` is already part of `llmTabBusy` (`app-key-bindings.ts`), `/model` switches to the LLM tab so `editorFocus` is false and typed characters cannot leak into the chat editor, and the Ctrl+C note is accurate since `handleAppKey` runs first in `tui-app.tsx`. Typing is gated on `status === "ready"`, so a fetch settling after the user typed cannot desync the cursor. Four things worth a look before this lands.

**1. The React-key explanation looks wrong.** Model ids are unique within a single list, and a key repeating *across renders* is exactly what keys are for — that is not a duplicate-key situation. Switching to `key={`row-${i}`}` is harmless here (the rows are plain `Text`), but it is an index key, and the stated cause does not explain the artifact. The more likely culprit is Ink falling back to non-erasing rendering once the frame exceeds the terminal height. Could you confirm the actual cause, or drop that justification from the commit message so the next reader is not misled?

**2. The fixed-height window does not cover every transition.** The `loading` and `error` branches in `llm-panel-modals.tsx` still render a different number of lines, so the `loading -> ready` height change remains. And a 12-row window plus the filter line makes the modal taller, which matters most on the short terminals where the original artifact showed up.

**3. The wizard picker now behaves differently for the same keystroke.** `CompatChatModelStep` in `providers-wizard.tsx` (`PICK_WINDOW = 12`) treats typing as "enter an id by hand", while this modal now treats it as "filter". With 337-model catalogs the wizard has the same usability problem, and two visually identical pickers reacting differently to the same key is a trap. Follow-up is fine, but it should not be forgotten.

**4. The printable-key guard is too broad.** `input && input.length > 0 && !key.ctrl && !key.meta` also admits Tab (`\t`) and, depending on what Ink puts in `input`, left/right arrows and page-up/down — all typed into the filter as junk. Worth excluding special keys explicitly, or checking that the character is actually printable.

Minor: `trim()` in the selector means typed spaces show up in the filter line but do not affect the result, and once a filter is active there is no way back to the currently active model (the cursor only lands on it at load time).

Nothing blocking. 1 and 4 are small; 2 and 3 can be a follow-up.

sosidudku1:
Thanks for the careful read, all four points addressed in the latest commit.

1. You were right on both counts. Ids are unique within a render, so the key explanation was wrong and is gone. We read through Ink 7's renderer to confirm the real mechanism: the standard log-update path repaints by erasing the previous frame's line count and rewriting, and `shouldClearTerminalForFrame` in ink.js switches to a full clear-terminal rewrite once a frame outgrows the viewport, since `eraseLines` cannot reach rows that scrolled past the top. The artifact showed up exactly when the unwindowed 337-row catalog pushed frames into that regime. The code comment now states what we verified and no more.
2. Loading and error now pad to the same line count as the ready branch, so no transition changes the frame height, and long ids/errors are truncate-end so wrapping cannot either. The window is also sized from the tab's row budget now (3 to 12 rows), so short terminals get a smaller but still constant window. Covered by height-equality tests at two budgets.
3. Agreed on the wizard picker. A typeable filter for it is already written and being tested locally; it lands in the next series of model-UX PRs rather than widening this one.
4. Done. The guard now excludes Tab, Enter, Esc, arrows, PgUp/PgDn and backspace/delete by flag, and rejects C0 controls and DEL by code point before anything reaches the query. Tests cover Tab, PgDn (including a raw `[6~` input), arrows and control bytes. Also added a test pinning the trim behavior: whitespace stays visible in the filter line but is ignored when matching.

PR #86, opened 2026-08-10
feat(tools): resolve project names to paths via os.fs.locate_project

### LINKED ISSUE #77: Let the agent resolve project paths without the user spelling out full paths
## Summary

Users have to type full absolute paths (`e:/_raylib`) for the agent to find a project. The request is for the agent to know the machine's folder layout well enough that "check my raylib project" resolves on its own.

## User feedback (Discord)

> "please add routing"

> "the folder / the route / how do i explain this / the agent knows where to look / because it knows the routes of the computer"

Screenshot showed the user typing `can you check on e:/_raylib` with path autocomplete visible.

## Open questions

The two plausible designs differ a lot in scope and in privacy implications:

1. **Scan and remember.** The agent indexes drives once and keeps a map of project folders. Convenient, but scanning a user's whole disk unprompted is invasive and slow on large volumes.
2. **User-declared roots.** The user names a few working directories ("my projects live in `e:/` and `d:/dev`") and the agent resolves fuzzy names against those. Cheaper, predictable, no surprise disk crawl.

Option 2 looks like the better default, with an explicit opt-in scan for people who want it.

## Notes

Path autocomplete already exists in the TUI input, so part of the plumbing is there. The missing piece is the agent resolving a fuzzy project name during a turn, not the user completing a path while typing.

sosidudku1:
## Problem

Issue #77: users must type full absolute paths (`e:/_raylib`) before the agent can find a project; "check my raylib project" should resolve on its own. The issue's own analysis prefers user-declared roots over an unsolicited disk scan, and this is a privacy-first product, so whole-disk indexing is off the table.

## Change

New read-only tool `os.fs.locate_project { name, limit? }`. A pasted absolute path that exists is returned immediately (`direct-path`); otherwise the name is resolved against three bounded sources in priority order: (1) session cwd and its ancestors; (2) recent sessions' working dirs via the new column-only `SessionStore.listRecentWorkingDirs` (no payload deserialisation on a pure_read path that can fan out in a 16-call batch), stat-checked so deleted dirs never surface; (3) the new `projects.roots` config list (v36, default `[]`) and its direct children, one level only, hidden dirs / `node_modules` / `__pycache__` / symlinks skipped, eligible dirs filtered before the 500-entry cap so loose files cannot evict project dirs.

Matching is NFC-normalized and case-insensitive with three tiers (exact, prefix, substring), so `raylib` finds `e:/_raylib`; backslash input parses on every host. Candidates are realpath-canonicalized before dedupe so symlink aliases never read as false ambiguity.

Result contract: one best-tier match returns the path plus weaker matches; ties return an explicit ambiguous result instructing the model to ask, computed on the full list so a tight `limit` cannot fake confidence; a miss is honest, reports how many roots were actually scanned (never the configured count), and repeats skipped/unreadable/truncated roots as note lines in the summary, because the model only ever sees the summary.

The descriptor teaches segment-only usage: pass a short folder-name segment (`raylib`), never the whole sentence; JSON examples render into the stable prefix so small local models copy a valid call shape. Root scans stream via `opendir` and stop at 500 eligible dirs; loose files, hidden names, and dependency caches never consume the cap. The tool honors `ToolContext.signal` (abort between roots, entries, and fs batches). Registered in `registerOsTools` with the other OS tools; bootstrap supplies the column-only session projection.

Wiring follows the AGENTS.md new-tool contract (resource class, args schema, frequent descriptor, GBNF branch); config migrates v35 to v36 transparently.

## Deliberate boundaries

No opt-in whole-disk scan (the issue lists it as an optional extra; follow-up on real demand). No fuzzy-distance matching, no per-root depth knob. README untouched (it documents no individual tools). Adding the descriptor and grammar branch changes the stable-prefix bytes once, so live llama-server slots take a one-time KV-cache invalidation right after the upgrade, the same shape as the MCP live add/remove precedent documented in AGENTS.md.

## Testing

35 new tests on the branch: 16 on candidate collection (each source; hidden/`node_modules`/`__pycache__`; one-level depth; the streamed 500-dir cap with files never consuming it; symlink alias dedupe and no-follow, EPERM-skipped on Windows; deleted session dirs; NFC/NFD; recency ordering), 14 on verdicts and the tool interface (the issue's `_raylib` example; exact-beats-substring; ambiguity incl. under a clamped limit; direct-path hit and fall-through; backslash input; honest miss incl. all-roots-unreadable; abort via `ctx.signal`; the descriptor contract: segment-only summary and examples that parse against the args schema; pure_read + frequent pinning), 1 session-store projection test, 3 config migration tests, 1 grammar parse test, plus the extended `registerOsTools` surface test. `tsc` clean; full-suite delta against the rebased base is zero new failures (same pre-existing set, verified via `git stash`).

Closes #77


src/prompt/default-tool-descriptors-a.ts
```diff
@@ -76,6 +76,12 @@ export const DEFAULT_TOOL_DESCRIPTORS_A: readonly ToolDescriptor[] = [
     argsSchema:
       "{ pattern: string | string[], cwd?: string, path?: string, ignore?: string[], absolute?: boolean, limit?: number, sortByMtime?: boolean, nocase?: boolean }",
   },
+  {
+    name: "os.fs.locate_project",
+    summary:
+      "Resolve a project mentioned by name ('my raylib project', 'in tasks-board') to a directory path when the user gives no full path. Checks the session cwd + ancestors, recent session dirs, and configured projects.roots (one level; never a whole-disk scan). Single match: use the returned path. Multiple: ask the user to pick. None: ask for the full path.",
```
Ooooze: **Blocker:** this summary steers the model toward a non-working `name`.

Matching in `parseArgs` / `matchTier` compares the **entire** `name` (or its last path segment) against directory basenames — exact / prefix / substring. There is no phrase tokenization.

```ts
// name: "my raylib project"  → query "my raylib project"
// basename "_raylib".includes(query) === false  → miss
// name: "raylib"             → substring hit on "_raylib"
```

The parenthetical examples (`'my raylib project'`, `'in tasks-board'`) look like tool-arg values, so small models will copy them into `args.name` and get honest misses for the exact Discord / issue #77 case.

**Fix:**
1. Reword the summary to require the distinctive directory segment (e.g. `raylib`, `tasks-board`), not the full user sentence.
2. Add `examples: ['{"name":"raylib"}', '{"name":"tasks-board"}']` so the args shape is unambiguous in the stable prefix.

Suggested summary sketch:

> Resolve a project directory by a short name segment the user mentioned (e.g. `raylib` → `…/_raylib`). Pass only that segment (or a pasted absolute path), not the full sentence. Searches session cwd + ancestors, recent session dirs, and `projects.roots` (one level; never whole-disk). Single match: use it. Multiple: ask. None: ask for the full path.

Ooooze (changes requested):
## Blocker

Descriptor copy teaches the model to pass a natural-language phrase as `name`, but matching is basename-only against the whole arg (or its last path segment). `name: "my raylib project"` misses `_raylib`; only `name: "raylib"` works. On small local models this will produce systematic false misses for the exact issue #77 utterance.

Please fix the stable-prefix summary + add `examples` before merge. Other notes from review are non-blocking (unbounded `readdir` before the 500 cap, test file >300 lines, missing `ctx.signal`, late bootstrap registration).

sosidudku1:
All five points addressed in the new commits (branch also rebased onto current main).

**Blocker (phrase-style summary):** Fixed. The summary now teaches a short folder-name segment ("raylib finds .../_raylib"), explicitly says "never the whole sentence", and the descriptor ships `examples: ['{"name":"raylib"}', '{"name":"tasks-board"}']` rendered into the stable prefix (the `examples` field already existed on `ToolDescriptor`). The runtime `description` says the same. Pinned by a descriptor contract test: no quoted multi-word value anywhere in the summary, and every example must parse as JSON and validate against the declared args schema (string `name`, no spaces, keys within `properties`, `required` includes `name`).

**Unbounded readdir:** Went further than an explanation: the scan now streams through `opendir` and stops pulling entries the moment the 501st eligible dir is seen, so a pathological root is never fully materialised (readdir has no partial mode, opendir does). Eligibility (dirs only, no hidden/`node_modules`/`__pycache__`) is checked in-stream before the cap, so loose files cannot evict project dirs. A mid-stream read error demotes the root to `unreadableRoots` and drops its partial candidates for determinism.

**Test file over 300 lines:** Split by theme into `fs-locate-project-sources.test.ts` (candidate collection) and `fs-locate-project.test.ts` (verdicts + tool interface), with shared fixtures in `fs-locate-project-test-helpers.ts` (same precedent as `git/test-helpers.ts`). Every source file is back under the 300-line rule.

**ctx.signal:** The tool now honors `ToolContext.signal`: checked at entry, between configured roots, between directory entries inside the opendir stream, and between the realpath/stat batches. Abort surfaces as a standard AbortError which `classifyFailure` maps to `cancelled`. Test: a pre-aborted signal rejects.

**Late registration:** Moved into `registerOsTools` next to the other OS tools. `SessionStore` has no construction-order dependencies, so bootstrap now constructs it right before the tool registry and passes the column-only projection through `RegisterOsToolsOptions.listRecentSessionDirs`; no lazy indirection, no temporal coupling. The `registerOsTools` surface test pins the tool's presence.

`tsc` clean; the tool area runs 72 tests green; the full suite shows zero new failures against the rebased base (same pre-existing set plus the two known wall-time flakes that also fail on the untouched baseline under identical load).

Ooooze (approved):
Blocker addressed in 85313be: segment-only descriptor + examples + contract test. Also covers the follow-ups (streamed root scan, abort signal, registerOsTools wiring, test split). LGTM.

Issue #106, opened 2026-08-11
OpenAI retry handling ignores Gemini error.details[].retryDelay metadata


yablokolabs:
## Summary

The OpenAI-compatible HTTP path now has bounded retries for 429/5xx responses and supports the standard `Retry-After` header. However, Gemini's OpenAI-compatible endpoint can return its cooldown only in structured error JSON, with no `Retry-After` header:

```json
{
  "error": {
    "code": 429,
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.RetryInfo",
        "retryDelay": "39s"
      }
    ]
  }
}
```

Atomic currently ignores this metadata. It retries using the 150/300 ms backoff and can exhaust all attempts before the provider's advertised cooldown has elapsed.

This behavior was observed on Atomic Agent v0.1.72. Generic cloud retries were added on `main` in `705e601`, so this issue is intentionally limited to the remaining provider-specific metadata gap rather than duplicating closed #13.

## Current behavior

[`httpErrorFromResponse`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/openai/openai-http.ts#L229-L242) reads the response body for an error preview but derives `retryAfterMs` only from the HTTP header.

[`resolveWaitMs`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/openai/openai-http.ts#L289-L316) therefore sees no provider cooldown and uses only the short exponential backoff.

## Expected behavior

For retryable 429/503 responses, Atomic should also recognize well-known structured retry metadata when `Retry-After` is absent. It should not burn all attempts before the provider-declared cooldown.

The retry remains bounded: if the declared delay exceeds the interactive retry budget, Atomic can surface a typed cooldown error instead of sleeping indefinitely.

## Suggested implementation

- Prefer a valid standard `Retry-After` header.
- Otherwise inspect bounded JSON error details for Google `RetryInfo.retryDelay` values such as `39s` or `1.5s`.
- Apply the existing retry cap/policy deliberately: either wait within budget or stop without issuing premature retries when the required delay is longer than that budget.
- Preserve caller cancellation and the existing rule that deterministic 4xx errors are not retried.

[...]

Issue #104, opened 2026-08-11
maxParallelToolCalls=1 still sends parallel_tool_calls: true to OpenAI-compatible providers


yablokolabs:
## Summary

`ATOMIC_AGENT_MAX_PARALLEL_TOOL_CALLS=1` limits Atomic's executor, but the main native-tools request path still asks OpenAI-compatible providers to generate parallel tool calls by sending `parallel_tool_calls: true`.

This prevents the setting from serving as a provider-compatibility control. It was a blocker with Gemini's OpenAI-compatible stream, where parallel calls lacked stable indices and were merged by the consumer.

Observed on Atomic Agent v0.1.72 and confirmed by inspection on v0.1.73/current `main` (`9d525ef4a004a9106608bea7a01ecd70ab279c2d`).

## Current behavior

The configured executor cap is loaded as `agent.maxParallelToolCalls`, but [`step-executor.ts`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/agent/step-executor.ts#L1113-L1122) builds the main native tool request with:

```ts
parallelToolCalls: true
```

[`openai-build-body.ts`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/openai/openai-build-body.ts#L20-L25) consequently serializes:

```ts
body.parallel_tool_calls = filtered.parallelToolCalls ?? true;
```

The provider capability model already exposes `supportsParallelTools`, but it is not reflected in this main request decision.

## Steps to reproduce

1. Configure an OpenAI-compatible provider with native tools.
2. Set `ATOMIC_AGENT_MAX_PARALLEL_TOOL_CALLS=1`.
3. Capture the `/chat/completions` request body for a normal agent step.
4. Observe `parallel_tool_calls: true`.

## Expected behavior

Atomic should send `parallel_tool_calls: false` when either:

- `agent.maxParallelToolCalls` is `1`; or
- the selected provider reports `supportsParallelTools: false`.

When the cap is greater than one and the provider supports parallel tools, the current parallel behavior can remain enabled.

## Suggested implementation

Derive `CompletionRequest.parallelToolCalls` from both runtime configuration and provider capabilities instead of hardcoding it in the step executor.

[...]

Issue #103, opened 2026-08-11
OpenAI SSE tool calls without index are merged when each event has a distinct id


yablokolabs:
## Summary

OpenAI-compatible streaming responses can omit `tool_calls[].index` while still supplying a stable `tool_calls[].id`. Atomic currently substitutes each SSE event's array position as the index. When a provider emits one distinct tool call per event, every call is therefore assigned index `0` and the stream consumer concatenates their names and arguments into one invalid call.

This was observed during a Gemini OpenAI-compatible run on Atomic Agent v0.1.72 and remains present by inspection on v0.1.73/current `main` (`9d525ef4a004a9106608bea7a01ecd70ab279c2d`).

## Current behavior

[`parse-sse-chunk.ts`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/openai/parse-sse-chunk.ts#L98-L114) falls back to the tool call's position within the current event:

```ts
index: typeof toolCall.index === "number" ? toolCall.index : fallbackIndex
```

[`openai-stream-consumer.ts`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/openai/openai-stream-consumer.ts#L113-L128) then reconstructs calls in a `Map<number, MutableToolCall>` keyed only by that index.

For two events shaped like this:

```text
data: {"choices":[{"delta":{"tool_calls":[{"id":"call_write","type":"function","function":{"name":"os__fs__write","arguments":"{...}"}}]}}]}

data: {"choices":[{"delta":{"tool_calls":[{"id":"call_read","type":"function","function":{"name":"os__fs__read","arguments":"{...}"}}]}}]}
```

both calls receive fallback index `0`. Observed failures included merged names such as:

```text
os.fs.writeos.fs.writeos.fs.writeos.fs.writeos.fs.write
```

## Expected behavior

When an SSE tool-call delta has no numeric `index` but has an `id`, Atomic should assign and retain a request-scoped stable index for that ID. Distinct IDs must remain distinct calls across separate events.

## Suggested implementation

Keep an ID-to-index map in the stream consumer:

1. Preserve an explicit numeric index when supplied.
2. Bind an accompanying ID to that explicit index.
3. If `index` is absent and the ID has been seen, reuse its assigned index.
[...]

PR #119, opened 2026-08-12
feat(llm): add first-class Gemini provider

### LINKED ISSUE #108: Add a first-class Gemini provider for Google's OpenAI-compatible endpoint
## Summary

Atomic's generic OpenAI-compatible provider always appends `/v1/...` to its configured base URL. Google's OpenAI-compatible API uses a versioned root ending in `/v1beta/openai`, so there is no straightforward direct Gemini configuration without a translating proxy.

This was finding §1 in Yabloko Labs' multi-provider evaluation. It is separate from the Gemini SSE, parallel-call, and retry defects already tracked in #103, #104, and #106.

## Current behavior

Current main registers `openai-compatible`, `openrouter`, and `aimlapi`, but no Gemini provider:

- [`register-built-in-providers.ts`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/registry/register-built-in-providers.ts#L40-L88)

The OpenAI base normalizer only strips a trailing `/v1`, while request methods append `/v1/chat/completions` or `/v1/models`:

- [`normalize-openai-base-url.ts`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/openai/normalize-openai-base-url.ts#L1-L10)
- [`openai-provider.ts`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/openai/openai-provider.ts#L85-L103)

Google's [current OpenAI-compatibility documentation](https://ai.google.dev/gemini-api/docs/openai) specifies this API root:

```text
https://generativelanguage.googleapis.com/v1beta/openai
```

With the generic Atomic path builder, that root receives another `/v1` segment instead of targeting Gemini's documented `/v1beta/openai/chat/completions` route.

## Proposed behavior

Add either:

1. a first-class `gemini` provider/preset that owns Gemini's endpoint paths and key convention; or
2. an endpoint-path option for OpenAI-compatible providers that preserves a versioned API root instead of unconditionally adding `/v1`.

A first-class provider would give the project a clean place for Gemini-specific retry metadata and stream normalization without making the generic provider conditional on hostnames.

## Related tracker history

#69 (implemented) and its duplicate #75 added presets for common endpoints that follow Atomic's normal `/v1` convention. Their candidate list did not include Gemini, and a preset alone cannot fix the extra `/v1` currently appended after Gemini's `/v1beta/openai` root. This request is therefore about endpoint-path behavior as well as setup convenience.

## Acceptance criteria

- Gemini can be configured directly without a localhost reverse proxy.
- The models and chat-completions requests use valid `/v1beta/openai/...` paths.
- Provider setup supports the Gemini API-key environment variable without writing the key to logs.
- Existing OpenAI-compatible, OpenRouter, and AIMLAPI paths remain unchanged.
- Tests pin URL construction for a base root that already contains `/v1beta/openai`.
- #103, #104, and #106 remain the canonical issues for SSE reconstruction, parallel-call control, and structured retry delay respectively.

yablokolabs:
## Problem

Atomic's generic OpenAI-compatible provider appends `/v1/...`, while Google's documented OpenAI-compatible API root is `/v1beta/openai`. That made direct Gemini setup require a translating proxy.

## Change

- Add a first-class `gemini` provider backed by Atomic's existing OpenAI transport.
- Route chat completions and model discovery through Google's documented `/v1beta/openai/...` paths.
- Resolve Gemini credentials from `GEMINI_API_KEY` without including the key in logs or errors.
- Add Gemini to the provider setup flow with a stable default chat model and no manual base-URL step.
- Preserve the existing OpenAI-compatible, OpenRouter, and AIMLAPI paths unchanged.

## Scope

This PR fixes native provider setup and endpoint routing only. Gemini SSE reconstruction, parallel-tool-call control, and structured retry delays remain tracked by #103, #104, and #106 respectively.

## Testing

- `npm run lint`
- Post-review Gemini export/configure regressions: 5 passed, 0 failed
- Affected provider/config/TUI suite: 389 passed, 0 failed
- `npm run build`
- Full suite: 3,953 passed, 6 failed, 4 skipped; no new failure names were introduced versus the captured upstream baseline (one baseline failure did not reproduce)
- `git diff --check`
- Public diff credential/private-path audit

Closes #108

Implementation and test preparation were LLM-assisted, then independently reviewed by Yabloko Labs.

### DIFF
--- a/src/tui/components/providers-wizard.tsx
+++ b/src/tui/components/providers-wizard.tsx
@@ -5,6 +5,7 @@ import {
   refreshAimlapiChatCatalogFromApi,
 } from "../../llm/provider/aimlapi/fetch-aimlapi-chat-catalog.js";
 import { fetchOpenAiCompatModels } from "../../llm/provider/openai/fetch-openai-compat-models.js";
+import { fetchGeminiModels } from "../../llm/provider/gemini/fetch-gemini-models.js";
 import {
   getCachedOpenRouterChatPicks,
   refreshOpenRouterChatCatalogFromApi,
@@ -142,17 +143,24 @@ function CompatChatModelStep(props: {
   const w = props.wizard;
   const baseUrl = baseUrlForWizard(w);
   const isCompat = w.kind === "openai-compatible";
+  const isGemini = w.kind === "gemini";
+  const canList = isCompat || isGemini;
   const [status, setStatus] = useState<{ loading: boolean; error: string | null }>(
-    { loading: isCompat, error: null },
+    { loading: canList, error: null },
   );
 
   useEffect(() => {
-    // Only the openai-compatible kind carries an operator-supplied base URL;
-    // any other kind would fire this at the default host with a stray key.
-    if (!isCompat) return;
+    // Only these kinds have a live model surface worth listing: openai-compatible
+    // carries an operator-supplied base URL, gemini has a fixed host keyed by its
+    // own key. Any other kind would fire at the default host with a stray key.
+    if (!canList) return;
     let alive = true;
     setStatus({ loading: true, error: null });
-    fetchOpenAiCompatModels(baseUrl, apiKeyForWizard(w)).then(
+    const apiKey = apiKeyForWizard(w);
+    const fetchModels = isGemini
+      ? fetchGeminiModels(apiKey)
+      : fetchOpenAiCompatModels(baseUrl, apiKey);
+    fetchModels.then(
       () => {
         if (alive) setStatus({ loading: false, error: null });
       },
@@ -170,12 +178,15 @@ function CompatChatModelStep(props: {
     };
     // Re-fetch only when the server changes; the key is fixed for this wizard run.
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [baseUrl, isCompat]);
+  }, [baseUrl, isCompat, isGemini]);
 
   const picks = listCompatChatModelPicks(w);
   if (picks.length > 0) {
+    const source = isGemini
+      ? "from Gemini /v1beta/openai/models"
+      : `from ${baseUrl}/v1/models`;
     return renderPickList({
-      title: `Chat model — ${picks.length} from ${baseUrl}/v1/models`,
+      title: `Chat model — ${picks.length} ${source}`,
       options: picks.map((id) => ({ label: id })),
       cursor: w.cursor,
       moveHint: "↑/↓ move",
@@ -184,10 +195,12 @@ function CompatChatModelStep(props: {
     });
   }
 
-  const hint = !isCompat
+  const hint = !canList
     ? "Enter to save · Esc back"
     : status.loading
-      ? `listing models from ${baseUrl}/v1/models…`
+      ? isGemini
+        ? "listing models from Gemini…"
+        : `listing models from ${baseUrl}/v1/models…`
       : status.error
       ? `${explainModelListError(status.error, w)} · type the id · Enter to save`
       : "Enter to save · Backspace to empty for the model list · Esc back";
--- a/src/tui/components/providers-wizard.tsx
+++ b/src/tui/components/providers-wizard.tsx
@@ -22,6 +22,7 @@ import {
   type ProvidersWizardKindRow,
 } from "../providers/providers-wizard-phases.js";
 import {
+  GEMINI_DEFAULT_CHAT_MODEL,
   listAimlapiEmbeddingModels,
   listOpenRouterEmbeddingModels,
   OPENAI_COMPAT_DEFAULT_BASE_URL,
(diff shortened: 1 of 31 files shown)


src/tui/providers/providers-orchestrator.ts
```diff
@@ -177,7 +178,10 @@ export class ProvidersOrchestrator {
       const apiKey = resolveLlmProviderApiKey(provider) ?? undefined;
       // Warm 1h cache resolves without a network round-trip, so the
       // loading state is only ever visible on a genuinely cold fetch.
-      const models = await fetchOpenAiCompatModels(baseUrl, apiKey);
+      const models =
+        provider.kind === "gemini"
+          ? await fetchGeminiModels(apiKey)
```
Ooooze, line 183: **Medium:** this correctly warms `fetchGeminiModels`, but the Cloud-pane renderer in `src/tui/llm-panel/llm-panel-row-builders.ts` → `inlineModelsForProvider` was not updated.

That helper only falls back to `getCachedOpenAiCompatModelsForBaseUrl(provider.baseUrl)`. Gemini entries have no `baseUrl`, so when `providersPanel.inlineModels` is null or belongs to another provider the secondary lookup always misses — and an empty `chatModel` can surface `OPENAI_COMPAT_DEFAULT_CHAT_MODEL` (`gpt-5.4-mini`).

Please add a gemini branch that reads `getCachedGeminiModels()` (same pattern as openai-compat’s module-cache fallback), and prefer `GEMINI_DEFAULT_CHAT_MODEL` over the openai-compat placeholder when `kind === "gemini"`.

src/tui/components/providers-wizard.tsx
```diff
@@ -191,7 +194,10 @@ function CompatChatModelStep(props: {
   return renderLineField({
     title: "Chat model id",
     value: w.chatModelLine,
-    placeholder: OPENAI_COMPAT_DEFAULT_CHAT_MODEL,
+    placeholder:
+      w.kind === "gemini"
```
Ooooze, line 211: **Medium:** placeholder for Gemini is good, but this step still has no live model list for `kind === "gemini"`.

`CompatChatModelStep` only fetches when `isCompat` (`openai-compatible`). Gemini skips `base_url` and lands here as free-text, so onboarding forces the operator to know a model id; `fetchGeminiModels` only warms later in the Cloud pane.

Please either:
1. also call `fetchGeminiModels(apiKey)` when `w.kind === "gemini"` and surface picks like the compat path, or
2. ship a short curated Gemini pick list + free-text escape.

Ooooze (changes requested):
## Code review — medium findings only

Architecture (thin `GeminiProvider` over OpenAi + `/v1beta/openai`) looks right. Blocking on two medium TUI gaps before merge.

| Severity | Where | Issue |
|---|---|---|
| **Medium** | `llm-panel-row-builders.ts` `inlineModelsForProvider` (not updated in this PR) | Gemini warms `fetchGeminiModels` cache, but the Cloud pane consumer still only reads `inlineModels` + `getCachedOpenAiCompatModelsForBaseUrl`. No `baseUrl` on gemini entries → secondary cache miss; if `inlineModels` is null/for another provider, list degrades (and empty `chatModel` can fall through to `OPENAI_COMPAT_DEFAULT_CHAT_MODEL` / `gpt-5.4-mini`). Add a `getCachedGeminiModels()` branch symmetric with openai-compat. |
| **Medium** | `providers-wizard.tsx` `CompatChatModelStep` | Gemini lands on free-text `chat_model_line` with no live `/v1beta/openai/models` fetch (`isCompat` gates openai-compatible only). Operator must guess the model id at setup; Cloud pane only warms later. Wire `fetchGeminiModels` here (or a short curated pick list + free-text escape). |

Low/nits left out of this review (`GOOGLE_API_KEY` fallback, parallel-tools default vs #104, trailing newlines, `KIND_ROW_ORDER` comment).

src/tui/providers/providers-orchestrator.ts
yablokolabs: Fixed in 6e6daf1. `inlineModelsForProvider` now has a gemini branch: it reads the gemini-keyed cache via a new `getCachedGeminiModelsForPanel()` (key-agnostic read mirroring `getCachedOpenAiCompatModelsForBaseUrl`) and falls back to `GEMINI_DEFAULT_CHAT_MODEL` — never the openai-compat placeholder. Added a regression test asserting an empty `chatModel` can't surface `gpt-5.4-mini`.

src/tui/components/providers-wizard.tsx
yablokolabs: Fixed in 6e6daf1. `CompatChatModelStep` now calls `fetchGeminiModels(apiKey)` for `kind === "gemini"` and surfaces the live `/v1beta/openai/models` list as a pick list, matching the openai-compatible path (`listCompatChatModelPicks` gained a gemini branch).

yablokolabs:
Both medium findings addressed in `6e6daf1` (pushed). Cloud pane gemini branch + wizard live fetch, with regression tests. tsc clean; affected suites 55/55 green. (Full-suite run shows pre-existing failures in memory/http/import areas on the base commit — unrelated to this PR.) Ready for re-review.

PR #120, opened 2026-08-12
fix(llm): adapt Qwen tagged tool calls

### LINKED ISSUE #105: OpenAI-compatible Qwen tagged tool calls are treated as assistant text
## Summary

Some Qwen deployments behind OpenAI-compatible endpoints return tool calls as tagged text instead of native `message.tool_calls`, for example:

```text
<tool_call>
<function=os.fs.read>
<parameter=path>/work/README.md</parameter>
</function>
</tool_call>
```

The tagged call may appear in `message.content` or `message.reasoning_content`. Atomic currently treats it as assistant text because the OpenAI response adapters only consume native `message.tool_calls`.

A second compatibility mismatch is that Atomic advertises flattened wire names such as `os__fs__read`, while the model can emit the human-facing dotted name `os.fs.read`.

This blocked a Tinker-hosted Qwen run on Atomic Agent v0.1.72 and remains reproducible by inspection on v0.1.73/current `main` (`9d525ef4a004a9106608bea7a01ecd70ab279c2d`).

## Current behavior

[`openai-normalise-response.ts`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/openai/openai-normalise-response.ts#L7-L20) reads native `message.tool_calls`, plain content, and `reasoning_content`, but does not recognize tagged tool syntax.

[`openai-tool-call-adapter.ts`](https://github.com/AtomicBot-ai/atomic-agent/blob/9d525ef4a004a9106608bea7a01ecd70ab279c2d/src/llm/provider/openai/openai-tool-call-adapter.ts#L78-L97) advertises escaped tool names and only normalizes names after a native tool call has already been reconstructed.

Without translation, the tagged call is interpreted as ordinary assistant content or routed to `reply` instead of executing the requested tool.

## Expected behavior

When no native `message.tool_calls` are present, a provider adapter should optionally recognize valid tagged calls from `content` or `reasoning_content`, convert them to native calls, and canonicalize names against the tools actually offered in the request.

Both of these should resolve to the advertised `os__fs__read` tool:

```text
os__fs__read
os.fs.read
```

## Suggested implementation

Add a bounded compatibility parser at the OpenAI provider-adapter layer that:

- activates only when native tool calls are absent;
- parses the known `<tool_call><function=...><parameter=...>` format;
- accepts only names corresponding to tools supplied in the request;
- maps dotted canonical names to their escaped wire names;
- coerces parameter values using the supplied JSON schema;
- passes the result through Atomic's normal tool validation and execution path.

A localhost adapter implementing this behavior enabled a meaningful 50-step Tinker/Qwen coding run.

## Acceptance criteria

- Tagged calls are parsed from both `content` and `reasoning_content`.
- Dotted and escaped names map to the same offered tool.
- String, integer, number, boolean, array, object, and null parameters are schema-coerced correctly.
- Multiple tagged calls preserve order and receive distinct indices in streamed output.
- Unknown tool names and malformed tagged text are not executed.
- Native OpenAI `message.tool_calls` remain the preferred path and behave unchanged.
- Ordinary prose containing angle brackets is not misclassified as a tool call.

yablokolabs:
## Problem

Some Qwen deployments behind OpenAI-compatible endpoints return tool calls as tagged text in `content` or `reasoning_content`, rather than native `message.tool_calls`. Atomic therefore treated valid calls as assistant prose. The model may also emit dotted names such as `os.fs.read` while Atomic advertises `os__fs__read`.

## Change

- Add an explicit opt-in OpenAI-compatible Qwen adapter instead of changing every OpenAI provider globally.
- Prefer native `message.tool_calls`; parse tagged calls only when native calls are absent.
- Resolve dotted and escaped names only against tools actually offered in the request.
- Coerce and validate tagged parameter values against the offered JSON schema, including nested arrays/objects, unions, enums, required fields, and additional-property rules.
- Preserve multiple-call order and distinct IDs through `completeStream()`; Atomic's existing executor then assigns deterministic batch indices.
- Fail closed for malformed tags, unknown tools or parameters, duplicate or missing required parameters, prototype-named edge cases, and invalid values; ordinary prose remains prose.

## Scope

The adapter is provider-specific and does not change generic OpenAI, OpenRouter, AIMLAPI, or Gemini behavior. Generic SSE reconstruction remains tracked separately by #103.

## Testing

- `npm run lint`
- Focused adapter/provider suite: 14 passed, 0 failed
- Affected provider/config/TUI suite: 392 passed, 0 failed
- `npm run build`
- Full suite: 3,956 passed, 6 failed, 4 skipped; all six failure names exactly match the captured upstream baseline (one upstream baseline failure — LlmHealthPoller re-probe — no longer reproduces) and no new failures were introduced
- `git diff --check`
- Public diff credential/private-path audit

Closes #105

Implementation and test preparation were LLM-assisted, then independently reviewed by Yabloko Labs.

### DIFF
--- a/src/llm/provider/openai/json-schema-support.ts
+++ b/src/llm/provider/openai/json-schema-support.ts
@@ -1,0 +1,129 @@
+type JsonSchema = Record<string, unknown>;
+
+const SUPPORTED_TYPES = new Set([
+  "string",
+  "integer",
+  "number",
+  "boolean",
+  "array",
+  "object",
+  "null",
+]);
+
+const SUPPORTED_KEYWORDS = new Set([
+  "$anchor",
+  "$comment",
+  "$defs",
+  "$id",
+  "$schema",
+  "additionalProperties",
+  "allOf",
+  "anyOf",
+  "const",
+  "contentEncoding",
+  "contentMediaType",
+  "default",
+  "deprecated",
+  "description",
+  "enum",
+  "examples",
+  "exclusiveMaximum",
+  "exclusiveMinimum",
+  "items",
+  "maxItems",
+  "maxLength",
+  "maxProperties",
+  "maximum",
+  "minItems",
+  "minLength",
+  "minProperties",
+  "minimum",
+  "multipleOf",
+  "not",
+  "oneOf",
+  "pattern",
+  "properties",
+  "readOnly",
+  "required",
+  "title",
+  "type",
+  "uniqueItems",
+  "writeOnly",
+]);
+
+export function assertSupportedJsonSchema(schema: JsonSchema): void {
+  for (const key of Object.keys(schema)) {
+    if (!SUPPORTED_KEYWORDS.has(key)) throw new Error(`unsupported schema keyword: ${key}`);
+  }
+  const types = typeof schema.type === "string" ? [schema.type] : schema.type;
+  if (types !== undefined) {
+    if (!Array.isArray(types) || types.length === 0) throw new Error("invalid schema type");
+    for (const type of types) {
+      if (typeof type !== "string" || !SUPPORTED_TYPES.has(type)) {
+        throw new Error("unsupported schema type");
+      }
+    }
+  }
+  for (const keyword of ["allOf", "anyOf", "oneOf"] as const) {
+    const branches = schema[keyword];
+    if (branches !== undefined) {
+      if (!Array.isArray(branches) || branches.length === 0) {
+        throw new Error(`invalid ${keyword}`);
+      }
+      for (const branch of branches) assertSupportedJsonSchema(asSchema(branch));
+    }
+  }
+  if (schema.not !== undefined) assertSupportedJsonSchema(asSchema(schema.not));
+  if (schema.items !== undefined) assertSupportedJsonSchema(asSchema(schema.items));
+  if (
+    schema.additionalProperties !== undefined &&
+    typeof schema.additionalProperties !== "boolean"
+  ) {
+    assertSupportedJsonSchema(asSchema(schema.additionalProperties));
+  }
+  for (const container of [schema.properties, schema.$defs] as unknown[]) {
+    if (container === undefined) continue;
+    const entries = asSchema(container);
+    for (const child of Object.values(entries)) {
+      assertSupportedJsonSchema(asSchema(child));
+    }
+  }
+  if (schema.enum !== undefined && (!Array.isArray(schema.enum) || schema.enum.length === 0)) {
+    throw new Error("invalid enum");
+  }
+  if (schema.required !== undefined && !Array.isArray(schema.required)) {
+    throw new Error("invalid required");
+  }
+  if (schema.pattern !== undefined) {
+    if (typeof schema.pattern !== "string") throw new Error("invalid pattern");
+    new RegExp(schema.pattern, "u");
+  }
+  for (const keyword of [
+    "exclusiveMaximum",
+    "exclusiveMinimum",
+    "maxItems",
+    "maxLength",
+    "maxProperties",
+    "maximum",
+    "minItems",
+    "minLength",
+    "minProperties",
+    "minimum",
+    "multipleOf",
+  ] as const) {
+    const assertion = schema[keyword];
+    if (assertion !== undefined && (typeof assertion !== "number" || !Number.isFinite(assertion))) {
+      throw new Error(`invalid ${keyword}`);
+    }
+  }
+  if (typeof schema.multipleOf === "number" && schema.multipleOf <= 0) {
+    throw new Error("invalid multipleOf");
+  }
+}
+
+function asSchema(value: unknown): JsonSchema {
+  if (value === null || typeof value !== "object" || Array.isArray(value)) {
+    throw new Error("invalid schema");
+  }
+  return value as JsonSchema;
+}
--- a/src/llm/provider/openai/openai-provider.ts
+++ b/src/llm/provider/openai/openai-provider.ts
@@ -28,7 +28,7 @@ import {
 import { normaliseOpenAiChatResponse } from "./openai-normalise-response.js";
 import { normalizeOpenAiBaseUrl } from "./normalize-openai-base-url.js";
 import { describeImageViaOpenAi } from "./openai-describe-image.js";
-import { adaptQwenTaggedToolResponse } from "./qwen-tagged-tool-response-adapter.js";
+import { adaptQwenCompletionResult, adaptQwenTaggedToolResponse } from "./qwen-tagged-tool-response-adapter.js";
 
 export interface OpenAiProviderOptions {
   id: string;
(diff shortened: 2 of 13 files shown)


src/llm/provider/registry/register-built-in-providers.ts
```diff
@@ -56,6 +56,26 @@ export function registerBuiltInProviderKinds(): void {
     });
   });
 
+  registerProviderKind("qwen-openai-compatible", (ctx) => {
```
Ooooze, line 63: **Medium — discoverability.** Registering `qwen-openai-compatible` fixes the runtime path, but the TUI Providers wizard / `KIND_ROW_ORDER` still only surfaces `openai-compatible`. Operators reproducing #105 (Tinker/Qwen behind an OpenAI-compatible URL) will keep the generic kind and never hit `taggedToolCompatibility: "qwen"`.

Please either:
- add a wizard/preset row for this kind, or
- document clearly (README / PR ops note) that `kind: "qwen-openai-compatible"` is required — `openai-compatible` alone does not enable the adapter.

Ooooze (commented):
## Review (medium findings)

Solid fix for #105: opt-in `qwen-openai-compatible`, native-first, fail-closed tagged parse, strong adapter tests. **Approve-leaning** — these are medium follow-ups, not merge blockers by themselves.

### Medium
1. **Discoverability** — kind is config/registry-only; TUI Providers wizard still only offers `openai-compatible`, so operators hitting #105 may never enable the adapter.
2. **Streaming disabled** — Qwen path forces unary `complete()` inside `completeStream()`; correct, but TUI/step stream UX goes silent until the full response.
3. **Content parse failure blocks reasoning** — malformed/`<tool_call>` noise in `content` returns early and never tries `reasoning_content`, which #105 says can also carry the call.
4. **Schema allowlist vs MCP** — unsupported keywords (e.g. `$ref`) fail closed and leave tagged calls as prose; default Atomic tools are fine, MCP `inputSchema` may silently regress into the original bug class.

Inline notes below. Happy to re-review if you want any of these in-tree before merge.

src/llm/provider/openai/openai-provider.ts
```diff
@@ -85,12 +89,19 @@ export class OpenAiProvider implements LlmProvider {
...
+      this.taggedToolCompatibility === "qwen"
+        ? adaptQwenTaggedToolResponse(json, request)
+        : json;
+    return normaliseOpenAiChatResponse(adapted, this.defaultChatModel);
   }
 
   async *completeStream(
     request: CompletionRequest,
   ): AsyncGenerator<StreamChunk, CompletionResult, void> {
+    if (this.taggedToolCompatibility === "qwen") {
```
Ooooze: **Medium — streaming effectively off.** Falling back to unary `complete()` is correct with `primeStream` / the fallback seam, and the provider test pins it. The main agent path still prefers `llmCompleteStream`, so TUI/step UI will show no deltas for this provider until the full response returns.

Fine as a v1 correctness tradeoff — please call it out in the PR/AGENTS note, and consider a follow-up that streams text normally and runs `adaptQwenTaggedToolResponse` on the final buffered message (same adapt seam, better UX).

src/llm/provider/openai/qwen-tagged-tool-response-adapter.ts
```diff
@@ -0,0 +1,181 @@
...
+    !message ||
+    (Array.isArray(message.tool_calls) && message.tool_calls.length > 0) ||
+    !request.tools?.length
+  ) {
+    return response;
+  }
+
+  const offered = indexOfferedTools(request.tools);
+  const contentCalls = parseSource(message.content, offered);
+  if (contentCalls === null) return response;
```
Ooooze: **Medium — malformed `content` blocks `reasoning_content`.** If `parseSource(content)` returns `null` (mixed prose + `<tool_call>`, truncated tags, etc.), we return the original response and never try `message.reasoning_content`.

#105 explicitly allows tagged calls in either field. Suggestion: when content is not a clean tagged-only payload (`null`), still attempt reasoning (or attempt reasoning whenever content yielded no calls), instead of fail-closed on the first field alone.

src/llm/provider/openai/json-schema-support.ts
```diff
@@ -0,0 +1,129 @@
...
+  "properties",
+  "readOnly",
+  "required",
+  "title",
+  "type",
+  "uniqueItems",
+  "writeOnly",
+]);
+
+export function assertSupportedJsonSchema(schema: JsonSchema): void {
```
Ooooze, line 54: **Medium — allowlist fail-closed vs MCP.** Rejecting unknown keywords (covered by the `$ref` test) is a clean safety default for Atomic's `DEFAULT_TOOL_ARGS_SCHEMAS`. MCP tools often ship `inputSchema` with `$ref` / draft-07 extras; those tagged calls will coerce-fail and remain assistant prose — the same user-visible failure mode as #105, just for MCP names.

Please document that MCP + tagged Qwen is unsupported for now, or soften: on unsupported schema keywords, fall back to a narrower coerce (e.g. JSON.parse / string passthrough) rather than dropping the whole tagged batch.

src/llm/provider/registry/register-built-in-providers.ts
yablokolabs: Documented in 5cf3c17 (README ops note): `kind: "qwen-openai-compatible"` is required to enable the adapter — plain `openai-compatible` deliberately does not. Keeping the kind config-only for now rather than adding a wizard row, since the runtime seam is the part #105 needs; happy to add a wizard/preset row as a follow-up if you'd like it discoverable in the TUI.

src/llm/provider/openai/openai-provider.ts
yablokolabs: Fixed in 5cf3c17 — went with your suggested follow-up rather than the doc note. The qwen path no longer falls back to unary `complete()`: it streams text/reasoning deltas live and runs the adapt seam on the fully-buffered final message (buffer-then-adapt). New `adaptQwenCompletionResult` rebuilds the wire envelope around the buffered `CompletionResult`. Test updated to an SSE fake asserting `stream: true`, live deltas, and adapted `tool_calls`.

src/llm/provider/openai/qwen-tagged-tool-response-adapter.ts
yablokolabs: Fixed in 5cf3c17. One correction to the framing: the code already fell through to `reasoning_content` when content was *empty* — the gap was only the `parseSource() === null` (malformed/noise) early-return. That path now also fails open to `reasoning_content`, so both empty and malformed content salvage the call. Regression test added.

src/llm/provider/openai/json-schema-support.ts
yablokolabs: Documented in 5cf3c17 (README limitation): the coercion supports a fixed JSON-Schema subset and rejects unknown keywords such as `$ref`, so MCP tools shipping draft-07 `inputSchema` with `$ref` fail coercion and remain prose — MCP + tagged Qwen is unsupported for now. A proper $ref resolver is a larger change; prefer to land it as a follow-up rather than soften fail-closed coercion here.

yablokolabs:
All four follow-ups addressed in `5cf3c17` (pushed): live buffered streaming (your suggested fix, not just a doc note), reasoning_content salvage on malformed content, and README ops notes for the `qwen-openai-compatible` kind requirement + the MCP `$ref` limitation. tsc clean; provider/registry/config suites 226/226 green. Happy to take the wizard row or a proper $ref resolver as follow-ups. Ready for re-review.

PR #124, opened 2026-08-14
feat: add /observe /manage /run tips to splash banner

Ayush7614:
## Summary
- Add splash-banner tips so `/observe`, `/manage`, and `/run` are discoverable from the empty chat screen.

## Test plan
- [x] Open TUI with an empty chat and confirm the splash banner lists `/observe`, `/manage`, and `/run`.
- [x] Related tests pass (`src/tui/components/chat-log.test.tsx` if covered).

### DIFF
--- a/src/tui/components/splash-banner.tsx
+++ b/src/tui/components/splash-banner.tsx
@@ -22,10 +22,13 @@ export function SplashBanner(): ReactElement {
         <Tip left="Enter" right="submit message to the agent" />
         <Tip left="/help" right="list all slash commands" />
         <Tip left="/sessions" right="switch to a previous thread" />
+        <Tip left="/observe" right="view diagnostics and world state" />
+        <Tip left="/manage" right="jump to the Manage tab" />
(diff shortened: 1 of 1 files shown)


sosidudku1:
Thanks for this, and sorry it sat for a while.

The idea is right — `/observe` and `/manage` genuinely were missing from the start page. Unfortunately the patch can't be merged as-is, for two reasons.

The splash screen was rewritten in #151 to adapt to terminal size. The tip list no longer lives in `splash-banner.tsx`; it moved to `SPLASH_TIPS` in `splash-fit.ts`, where each entry carries a short description as well as a full one, and the list is dropped from the tail when the window is small. That's what the conflict is — the lines here no longer have a place to go.

The other thing: `/run` isn't a command of its own, it's an alias for `/chat`, and it returns you to single-view chat mode rather than starting a new run. So that tip would have been misleading.

I've carried the `/observe` and `/manage` tips over in the new format, placed near the tail so a narrow terminal still shows the basics first. Closing this one, but thank you for spotting the gap.

PR #125, opened 2026-08-14
fix: correct Local AI-First Agent text expectation in test

Ayush7614:
## Summary
- Update the chat-log splash test to expect the current “Local AI-First Agent” title.
- Includes the splash-banner command tips from `fix/splash-banner-commands` (this branch is stacked on that one).

## Test plan
- [ ] `npm test -- src/tui/components/chat-log.test.tsx`
- [ ] Confirm the splash banner title and `/observe` `/manage` `/run` tips render in TUI.

### DIFF
--- a/src/tui/components/chat-log.test.tsx
+++ b/src/tui/components/chat-log.test.tsx
@@ -25,7 +25,7 @@ describe("ChatLog", () => {
     const state = createInitialTuiState(BASE_SESSION);
     const { lastFrame } = render(<ChatLog state={state} />);
     const text = strip(lastFrame() ?? "");
-    expect(text).toContain("Local-First AI Agent");
+    expect(text).toContain("Local AI-First Agent");
     expect(text).toContain("/help");
   });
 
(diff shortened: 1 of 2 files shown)


sosidudku1:
Thanks for this one, @Ayush7614 — and sorry it sat open as long as it did.

You caught a real inconsistency. That string was introduced in one commit that hand-copied the same tagline into four places, and three of them got `Local AI-First Agent` while `chat-log.test.tsx` got the words transposed. It sat there unnoticed for weeks. Spotting a two-word swap in a test assertion is exactly the kind of thing that is easy to scroll past, so thank you for actually stopping on it.

Since then `main` has moved past it in a way that makes the patch unmergeable:

1. **The tagline assertion is gone.** #151 made the start page adapt to terminal size, so the tagline is not always drawn — a short terminal drops it. The test now asserts that *some* logo mark is rendered instead of matching on text, and the string itself moved into a single `TAGLINE` constant in `logo.tsx`. One source, so the two copies can no longer drift apart.
2. **The splash tips moved.** They no longer live in `splash-banner.tsx`; the list is now `SPLASH_TIPS` in `splash-fit.ts`, where each entry carries both roomy and terse copy and the tail is dropped on narrow surfaces. The `<Tip>` elements this patch edits are not there anymore.

Closing as superseded rather than rejected — the underlying fix was right, we just arrived at it from a different direction while this was waiting. Three PRs from you in one day, all of them pointing at genuine problems, is a better hit rate than most drive-by contributions get.

If you feel like sticking around, our Discord is at https://discord.gg/kXWDFSJpMW. Given you do community work, you would probably find the conversation there interesting either way — and I would rather point you at something currently open than have you spend effort on another patch that history overtakes.

PR #126, opened 2026-08-14
fix: correct preset row index in provider wizard tests

Ayush7614:
## Summary
- Fix provider-wizard tests to use `3 +` preset row indexing (was `2 +`) so Local Ollama / Ollama Cloud rows are selected correctly.
- This branch is stacked: also includes splash-banner command tips and the Local AI-First Agent test-title fix.

## Test plan
- [x] `npm test -- src/tui/llm-panel/providers/provider-wizard.test.ts` (or the colocated wizard test that asserts preset rows)
- [x] Walk the provider setup wizard and confirm Local Ollama and Ollama Cloud land on the expected rows.

### DIFF
--- a/src/tui/components/chat-log.test.tsx
+++ b/src/tui/components/chat-log.test.tsx
@@ -25,7 +25,7 @@ describe("ChatLog", () => {
     const state = createInitialTuiState(BASE_SESSION);
     const { lastFrame } = render(<ChatLog state={state} />);
     const text = strip(lastFrame() ?? "");
-    expect(text).toContain("Local-First AI Agent");
+    expect(text).toContain("Local AI-First Agent");
     expect(text).toContain("/help");
   });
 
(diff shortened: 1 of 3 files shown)


sosidudku1:
Thanks for digging into this, @Ayush7614 — and sorry it took us a while to get back to you.

Since this was opened, `main` has moved past all three of the fixes here, so there is nothing left to merge:

1. **Preset row index.** The magic offset is gone entirely. `providers-wizard-key-bindings.test.ts` now derives rows from a `presetRowIndex()` helper that reads the real `KIND_ROW_ORDER`, so there is no `2 +` or `3 +` to correct. You were right that the hardcoded offset was fragile — we ended up removing the offset instead of updating it.
2. **Splash banner tips.** #151 made the start page adapt to terminal size. The tip list moved out of the component into a `SPLASH_TIPS` constant in `splash-fit.ts`, where each entry carries both roomy and terse copy and the tail is dropped on short terminals. The `<Tip>` elements this patch edits no longer exist.
3. **Chat log title.** That assertion was dropped. The test now checks that a logo is drawn at all, since the mark shrinks with the surface.

Closing as superseded, not rejected. The `/observe`, `/manage` and `/run` tips are a fair point on their own — `SPLASH_TIPS` is ordered by priority and the tail gets cut first, so it is worth a separate discussion about which entries earn a slot.

If you want to pick up something else, we hang out in Discord: https://discord.gg/kXWDFSJpMW — happy to point you at an issue that is actually open, or talk through what the wizard needs next.

PR #128, opened 2026-08-15
docs: highlight new Ollama (local) preset in README

yablokolabs:
Adds mention of dedicated Ollama (local) preset in provider wizard
- Notes auto-detection and streamlined setup in v0.2.1
- Improves discoverability of new local model preset feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,8 @@ In the TUI, open the LLM tab, add a provider, and pick **Ollama (local)** (or **
 
 Model ids are the tags the server reports, `qwen2.5:0.5b` or `llama3.2:latest` for Ollama, so use the same name you passed to `ollama pull`. The list comes from the server's own `/v1/models`, which means anything you have pulled shows up without a restart.
 
+**New in v0.2.1**: The provider wizard now includes a dedicated **Ollama (local)** preset that auto-detects your local Ollama server and provides a streamlined setup experience.
+
 | Preset | Endpoint |
 | --- | --- |
 | Ollama (local) | `http://localhost:11434` |


sosidudku1:
Thanks for taking the time to open this, and for reading the local-model section closely enough to want it clearer.

I'm going to pass on this one, for three reasons:

**It's already covered.** Two paragraphs above the insertion point, the section walks through exactly this: open the LLM tab, add a provider, pick **Ollama (local)**, and the wizard skips the key screen because a local server has no API key. The preset table right below the new line lists `Ollama (local) | http://localhost:11434`, and the paragraph after it separates the local preset from Ollama Cloud. The added sentence restates its neighbours rather than adding to them.

**The auto-detect claim isn't accurate.** The preset does not detect anything. It's a hardcoded default (`baseUrl: "http://localhost:11434"`, flagged `local`) in `src/tui/providers/provider-presets.ts`. There's no port probe and no check that a server is running — if `ollama serve` isn't up, you find out when the model list comes back empty. Shipping "auto-detects your local Ollama server" would set an expectation the code doesn't meet.

**We don't keep version badges in the README.** There are no other "New in vX" markers in the file, and the preset landed back in v0.2.1 (#122) — we're several releases past that now, so the badge would read as stale the day it merged. Release notes are where that belongs.

If you'd like to improve this section, here's a gap that's actually open: the README never says what happens when the Ollama server *isn't* running — the model list simply comes back empty, with no hint to go start it. A sentence covering that would be a real addition, and I'd be glad to review it.

yablokolabs:
Opened a new PR with the requested rework and the relevant updates consolidated into a clean follow-up: #204

PR #129, opened 2026-08-16
feat: add Sarvam AI provider support

yablokolabs:
## Summary

Adds first-class [Sarvam AI](https://www.sarvam.ai) provider support, routed through the existing OpenAI-compatible `OpenAiProvider` — the same pattern as every other compat provider.

## Changes

| File | Change |
|---|---|
| `src/config/llm-config.ts` | Add `sarvam` to `PROVIDER_KINDS` so `kind: "sarvam"` entries validate |
| `src/llm/provider/registry/register-built-in-providers.ts` | Register the `sarvam` provider kind (OpenAI-compatible transport, vision defaulted off, parallel tools on) |
| `src/tui/providers/provider-presets.ts` | Add the Sarvam preset (`https://api.sarvam.ai`, `SARVAM_API_KEY`) to the TUI provider wizard |

## Features

- Sarvam AI integration with OpenAI-compatible API (no custom transport needed)
- New TUI wizard preset: pick "Sarvam", paste `SARVAM_API_KEY`, done
- CLI/config support via `llm.providers[].kind = "sarvam"` or the openai-compatible flow
- Support for `sarvam-105b` and `sarvam-105b-conversations` (model list comes from the live `/v1/models` endpoint, per existing preset convention)
- `SARVAM_API_KEY` env-var authentication with the preset's own env var (no cross-service key collisions)
- Preserves xAI (Grok) and all existing providers — purely additive

## Verification

- `npm run lint` (tsc noEmit): passes
- `npm test`: 3982+ passed; the only failures are pre-existing on `main` (splash-banner, llm-health-poller, send-message-concurrency, persist-embedding-hybrid-recall) and unrelated to this change — verified on a clean `main` checkout
- `npm run build`: passes
- Preset ordering pinned by `provider-presets.test.ts` ("is sorted alphabetically by label") and `providers-wizard-key-bindings.test.ts`: Sarvam sits between "Ollama Cloud" and "Together AI"

## Notes

- Sarvam advertises an OpenAI-compatible API, so no dedicated provider class is needed; `supportsVision` defaults to `false` (Sarvam models are text-only today).

### DIFF
--- a/src/config/llm-config.ts
+++ b/src/config/llm-config.ts
@@ -49,6 +49,7 @@ const PROVIDER_KINDS = new Set([
   "openrouter",
   "aimlapi",
   "gemini",
+  "sarvam",
 ]);
 
 function parseProviderId(raw: unknown, field: string): string {
--- a/src/llm/provider/registry/register-built-in-providers.ts
+++ b/src/llm/provider/registry/register-built-in-providers.ts
@@ -124,4 +124,23 @@ export function registerBuiltInProviderKinds(): void {
       requestTimeoutMs: entry.requestTimeoutMs,
     });
   });
-}
+
+  registerProviderKind("sarvam", (ctx) => {
+      const entry = ctx.entry;
+      if (!entry.baseUrl || !entry.defaultChatModel) {
+        throw new Error(
+          `sarvam provider "${entry.id}" requires baseUrl and defaultChatModel`,
+        );
+      }
+      return new OpenAiProvider({
+        id: entry.id,
+        baseUrl: entry.baseUrl,
+        apiKey: entry.apiKey ?? "",
+        defaultChatModel: entry.defaultChatModel,
+        headers: entry.headers,
+        supportsVision: entry.supportsVision ?? false,
+        supportsParallelTools: entry.supportsTools ?? true,
+        requestTimeoutMs: entry.requestTimeoutMs,
+      });
+    });
+  }
--- a/src/tui/providers/provider-presets.ts
+++ b/src/tui/providers/provider-presets.ts
@@ -121,6 +121,13 @@ export const PROVIDER_PRESETS: readonly ProviderPreset[] = [
     note: "hosted Ollama, models listed without a key",
   },
   {
+    id: "sarvam",
+    label: "Sarvam",
+    baseUrl: "https://api.sarvam.ai",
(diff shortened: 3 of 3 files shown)


sosidudku1:
Hi @yablokolabs, thank you for putting this together — and for the detailed writeup.

We're going to pass on this one, but the reasoning is about our architecture rather than the quality of your work, so let me explain properly.

The provider presets in `src/tui/providers/provider-presets.ts` are deliberately *not* provider kinds. As the file header notes, every preset resolves to the existing `openai-compatible` kind with `baseUrl` filled in — that's how Cerebras, Groq, Mistral, Together AI, xAI and the rest are wired. Since Sarvam is OpenAI-compatible, registering a dedicated `sarvam` kind duplicates the `openai-compatible` factory almost line for line, and each such duplicate makes the registry harder to keep consistent as it grows.

Two smaller notes for context:

- The added block in `register-built-in-providers.ts` closes the function in the wrong place, so the new registration ends up nested inside the preceding one. Worth a second look at the build output.
- The file header also asks that each preset URL be verified live against `/v1/models` before it lands. That check isn't in the PR.

**The good news:** Sarvam already works today without any code changes. In the TUI provider wizard, pick the manual **OpenAI-compatible** entry, set the base URL to Sarvam's API root and paste your key — the model list comes from the server's own `/v1/models`, so nothing needs updating when Sarvam ships a new model. The equivalent config entry is `kind: "openai-compatible"`.

If you'd like Sarvam to appear as a named row in the wizard, we'd welcome a much smaller PR: a single entry in `PROVIDER_PRESETS` (alphabetically between "Ollama Cloud" and "Together AI"), no new provider kind, plus a note confirming the base URL answers `/v1/models`. That's about five lines and we'd be glad to review it.

Thanks again for the contribution and for the interest in the project.

yablokolabs:
hi [sosidudku1](https://github.com/sosidudku1) Thanks, that makes sense.

I understand the distinction now: this should be a preset-only addition rather than a new provider kind. I'll follow up with a much smaller PR that just adds the Sarvam preset entry and confirms the base URL responds to /v1/models.

yablokolabs:
Opened the smaller follow-up PR here: https://github.com/AtomicBot-ai/atomic-agent/pull/205

It contains just the preset entry and the /v1/models confirmation, per your guidance.

PR #151, opened 2026-08-19
feat(tui): start page adapts to terminal size

plombeer31:
## Problem

The TUI start page was built from three pieces that never looked at the terminal:

| Piece | Hard-coded size |
|---|---|
| Brand mark + `ATOMIC AGENT` wordmark (`logo.tsx`) | 34 cols art + 3 gap + 46 cols wordmark = **83 cols**, **20 rows** |
| Tip list (`splash-banner.tsx`) | 8 fixed rows, `padEnd(24)` + descriptions up to 62 chars ≈ **90 cols** |
| Right rail (`sidebar.tsx`) | `width={30}` flat, `MAX_SESSION_ROWS=10`, `MAX_TASK_ROWS=5`, height never consulted |

So the splash needed ~90 columns and ~29 rows, plus the 8 rows of prompt chrome — **37 rows** — before it could render honestly. Anything smaller broke, and as `row-window.ts` already notes, Ink 7 does not clip an over-tall frame: it overlaps earlier lines.

Two extra details made it worse than the raw numbers suggest:

- The rail appeared at ≥100 columns and took a flat 30, leaving **70** columns for a splash that wanted 87 — so the artwork wrapped at *every* width from 100 up to ~118.
- The rail had no `flexShrink={0}`, so Yoga let the chat column claw columns back out of it. On a 100-column terminal the rail rendered ~18 wide instead of 30 and truncated `(no sessions yet)` to `(no sessions ye…`.

`Logo` already had a `compact` prop; `SplashBanner` never passed it. `useTerminalSize()` already existed and was already used by `ChatLog` and `TuiApp`.

### Before — real `atomic-agent tui`, 80×24

```
  atomic-agent v0.2.2  |  ▸ Run  ·    Observe  ·    Manage
       -:::::::::::::-
    -::::::::::::::::-
  :::::::::::::::::::::::::::::::-
  ::::::::::::::::::::::::::::::::   ▄▀█ ▀█▀ █▀█ █▀▄▀█ █ █▀▀   ▄▀█ █▀▀ █▀▀ █▄ █
  ::::::::::::::::::::::::::::::::   █▀█  █  █▄█ █ ▀ █ █ █▄▄   █▀█ █▄█ ██▄ █ ▀█
  ::::::::::::::::::::::::::::::::
  -----------:::::::::::::::::---=   Local AI-First Agent
  @@@@@@@@@@*-::::::::::::-=+#%%@
            -:::::::::::-+#@
            -::::::::::=#@
      • /sessions               switch to a previous thread
      • /model                  change the chat model
      • /tasks                  jump to the Tasks tab (Option 4 cron + ingress
      • /import                 open the Import tab (one-shot Hermes ->
      • Ctrl+C ×2               quit (once aborts a running turn)
```

Top of the mark cut off, wordmark clipped mid-glyph, the `Enter` / `/help` / `/new` tips gone, two descriptions sliced mid-word.

### Before — 60×20

```
  atomic-agent v0.2.2  |  ▸ Run  ·    Observe  ·    Manage
  :::::::::::::::::::::-
  ::::::::::::::::::::::   ▄▀█ ▀█▀ █▀█ █▀▄▀█ █ █▀▀   ▄▀█ █▀▀
  ::::::::::::::::::::::   █▀█  █  █▄█ █ ▀ █ █ █▄▄   █▀█ █▄█
  ::::::::::::::::::::::
  -:::::::::::::::::---=   Local AI-First Agent
  *-::::::::::::-=+#%%@
  -:::  • /sessions               switch to a previous ds
  -:::  • /tasks                  jump to the Tasks tab
  -:::  • /import                 open the Import tab
  -:::  • Ctrl+C ×2               quit (once aborts a
```

Here the overlap is unmistakable — tip rows are painted straight over the artwork.

## What this does

Everything on the start page is now sized against the live terminal.

**`src/tui/layout.ts` (new)** — the shared geometry, React-free:

- `isSidebarVisible(columns)` / `computeSidebarWidth(columns)` — the rail is `clamp(24, 25% of columns, 34)` instead of a flat 30. The ≥100-column visibility gate is unchanged.
- `computeChatWidth(columns)` — what is left for the chat column once the root padding and the rail have taken theirs. `TuiApp` and `SplashBanner` read the same number, so the splash can no longer disagree with the rail about where the boundary is.
- `computeChatViewportRows(rows, columns?)` — the old `CHROME_ROWS` constant from `chat-log.tsx`, now shared, plus a correction for narrow terminals where the status bar, the hint strip and the prompt placeholder each wrap onto extra lines.
- `computeSidebarRowBudget(rows)` — splits the rail's usable height ~2:1 between Sessions and Tasks, both with a floor of 1 and the old 10 / 5 as caps.

**`src/tui/components/splash-fit.ts` (new)** — pure breakpoint maths. Takes the surface size, returns which mark to draw, whether the wordmark and tagline fit, how many tips survive, the tip-label column width, and whether descriptions are full, terse or dropped. Being React-free, it is unit-tested as a table rather than through rendered frames.

**`logo.tsx`** — the mark now comes in three sizes: `full` (the original 34×20 drawing, untouched), `small` (17×10) and `mini` (a single 14-column line). The artwork moved into an exported `LOGO_ART` record so a test can measure it. `compact` still works for existing callers.

**`splash-banner.tsx`** — measures the terminal, applies the fit plan. Tips live in one priority-ordered array with a long and a short description each; the tail is dropped first when rows run out, and descriptions collapse to the terse form before any tip disappears. Takes an optional explicit `size` so tests are deterministic.

**`sidebar.tsx`** — stays purely presentational: `width`, `maxSessionRows` and `maxTaskRows` arrive as props from `TuiApp`, which owns the measurement. Preview truncation is derived from the rail width instead of the hard-coded 28 / 24. Both panes now use the shared `computeRowWindow` from `row-window.ts` (replacing a private copy), so the Tasks pane finally scrolls to keep its cursor visible instead of silently slicing the first five rows. Added `flexShrink={0}` so the rail keeps the width it was given.

The mark has priority over the tip list: a window that grows tall enough for a bigger mark spends its new rows on the artwork, so the tip count can legitimately drop across a variant change. Within one variant the list only ever grows. That trade-off is documented and pinned by a test.

### After — 120×40

```
  atomic-agent v0.2.2  |  ▸ Run  ·    Observe  ·    Manage
                                                                                          │ Sessions
                 -:::::::--                                                               │ (no sessions yet)
                 -::::::::-                                                               │
                -:::::::::-                                                               │ Tasks
               -::::::::::-                                                               │ (no active tasks)
              -:::::::::::-                                                               │
            -:::::::::::::-                                                               │
         -::::::::::::::::-                                                               │
     -::::::::::::::::::::::::::::::::-                                                   │
     ::::::::::::::::::::::::::::::::::   ▄▀█ ▀█▀ █▀█ █▀▄▀█ █ █▀▀   ▄▀█ █▀▀ █▀▀ █▄ █ ▀█▀  │
     ::::::::::::::::::::::::::::::::::   █▀█  █  █▄█ █ ▀ █ █ █▄▄   █▀█ █▄█ ██▄ █ ▀█  █   │
     -:::::::::::::::::::::::::::::::::                                                   │
     =------------:::::::::::::::::---=   Local AI-First Agent                            │
      @@@@@@@@@@@*-::::::::::::-=+#%%@                                                    │
                 -:::::::::::-+#@                                                         │
                 -::::::::::=#@                                                           │
                 -:::::::::=#                                                             │
                 -::::::::-*                                                              │
                 -::::::::=                                                               │
                 +--------*                                                               │
                   %%%%%%                                                                 │
                                                                                          │
              • Enter                   submit message to the agent                       │
              • /help                   list all slash commands                           │
              • /sessions               switch to a previous thread                       │
              • /new                    start a fresh session                             │
              • /model                  change the chat model                             │
              • /tasks                  jump to the Tasks tab (cron + ingress UI)         │
              • /import                 open the Import tab (Hermes migration)            │
              • Ctrl+C ×2               quit (once aborts a running turn)                 │
                                                                                          │
                                                                                          │
  │                                                                                       │
  │ ❯  Type a message or `/` for commands…                                                │
  │                                                                                       │
  │ llama.cpp                                                                             │
  ╹                                                                                       │
  [enter] send [alt+enter] newline [tab] sidebar  [fn+↑↓] scroll [/] commands  [ctrl+c]   │
```

### After — 100×30

The rail border now sits exactly where `computeChatWidth(100)` says it should — column 75 — and `(no sessions yet)` is no longer truncated.

```
  atomic-agent v0.2.2  |  ▸ Run  ·    Observe  ·    Manage
                                                                           │ Sessions
                                    -:::--                                 │ (no sessions yet)
                                    -::::-                                 │
                                   -:::::-                                 │ Tasks
                                 -:::::::-                                 │ (no active tasks)
                              -:::::::::::::::-                            │
                              :::::::::::::::::                            │
                              =-----:::::::::-=                            │
                               @@@@@*-::::-=#%                             │
                                    -::::-*                                │
                                    +----*                                 │
                                                                           │
      • Enter                   submit message to the agent                │
      • /help                   list all slash commands                    │
      • /sessions               switch to a previous thread                │
      • /new                    start a fresh session                      │
      • /model                  change the chat model                      │
      • /tasks                  jump to the Tasks tab (cron + ingress UI)  │
      • /import                 open the Import tab (Hermes migration)     │
      • Ctrl+C ×2               quit (once aborts a running turn)          │
                                                                           │
  │                                                                        │
  │ ❯  Ask anything about your codebase…                                   │
  │                                                                        │
  │ llama.cpp                                                              │
  ╹                                                                        │
  [enter] sen[alt+enter]     [tab] sideba[fn+↑↓] scrol[/] commands[ctrl+c] │
```

### After — 60×20

```
  atomic-agent v0.2.2  |  ▸ Run  ·    Observe  ·    Manage
                        + ATOMIC AGENT

                    • Enter     send message
                    • /help     all commands
                    • /sessions past threads
                    • /new      new session
                    • /model    pick model
                    • /tasks    Tasks tab
                    • /import   Hermes import
                    • Ctrl+C ×2 quit

  │
  │ ❯  Ask anything about your codebase…
  │
  │ llama.cpp
  ╹
  [enter]  [alt+enter]  [tab]    [fn+↑↓]   [/]       [ctrl+c
  send  ·  newline  ·   sidebar  scroll  · commands  ] quit
```

### After — 50×18

```
  atomic-age        | ▸ Run  ·    Observe  ·
  t         v0.2.2    Manage
                   + ATOMIC AGENT

        • Enter     submit message to the agent
        • /help     list all slash commands
        • /sessions switch to a previous thread
        • /new      start a fresh session

  │
  │ ❯  Ask anything about your codebase…
  │
  │ llama.cpp
  ╹
```

## Tests

- `src/tui/layout.test.ts` — rail width clamps, chat-width monotonicity, row-budget floors and caps, viewport chrome.
- `src/tui/components/splash-fit.test.ts` — breakpoint table plus invariants swept over 10–200 columns × 2–60 rows: the planned layout always fits its surface, the mark never shrinks as the terminal widens, the tip list never shrinks within a variant.
- `src/tui/components/splash-fit.render.test.tsx` — renders the real component at 40×12, 60×20, 80×24, 100×30 and 100×50 and asserts no line exceeds the surface width and the frame never exceeds its rows, in the style of `manage-panel-fit.test.tsx`.
- `src/tui/components/logo-fit.test.ts` — measures the actual artwork rows against the `LOGO_METRICS` the breakpoints are built on, so the two can never drift apart silently.
- `src/tui/components/sidebar.test.tsx` — per-pane row budget, the Tasks pane scrolling to its cursor, and previews narrowing with the rail.

Two tests that were already failing on `main` are fixed on the way past: `splash-banner.test.tsx` asserted `/observe`, `/manage` and `/run` (the tips changed in #97), and `chat-log.test.tsx` asserted the banner text `Local-First AI Agent` when the component renders `Local AI-First Agent`.

## Verification

Node 25, `npm run build` clean. `npx vitest run src/tui`: the only remaining failures are the ones already failing on `main` in the same environment (`tui-app.test.tsx` nav-flakes under parallel load, `llm-health-poller`, `persist-embedding-hybrid-recall`). Frames above are real captures of `atomic-agent tui` driven through a PTY at each exact size and replayed through a terminal emulator.

Known limit: at 45×16 — narrower than any of the sizes above — Ink can still leave a stray character behind when the rotating prompt placeholder wraps and shifts the vertically-centred splash by a row. The frame no longer overlaps itself, which was the actual bug, but that size is past the point where the chrome alone fills the window.

### DIFF
--- a/src/tui/components/chat-log.test.tsx
+++ b/src/tui/components/chat-log.test.tsx
@@ -25,10 +25,11 @@ describe("ChatLog", () => {
     const state = createInitialTuiState(BASE_SESSION);
     const { lastFrame } = render(<ChatLog state={state} />);
     const text = strip(lastFrame() ?? "");
-    // The mark shrinks with the surface, so assert on the plus bar the
-    // artwork always keeps rather than on a wordmark that only a tall
-    // terminal earns. See `splash-fit.render.test.tsx`.
-    expect(text).toContain(":::");
+    // The mark shrinks with the surface — shaded art at full size, half
+    // block art below it — so assert that *some* mark is drawn rather
+    // than on a wordmark only a tall terminal earns. See
+    // `splash-fit.render.test.tsx`.
+    expect(text).toMatch(/:::|[█▀▄]/u);
     expect(text).toContain("/help");
   });
 
--- a/src/tui/components/chat-log.test.tsx
+++ b/src/tui/components/chat-log.test.tsx
@@ -25,7 +25,10 @@ describe("ChatLog", () => {
     const state = createInitialTuiState(BASE_SESSION);
     const { lastFrame } = render(<ChatLog state={state} />);
     const text = strip(lastFrame() ?? "");
-    expect(text).toContain("Local-First AI Agent");
+    // The mark shrinks with the surface, so assert on the plus bar the
+    // artwork always keeps rather than on a wordmark that only a tall
+    // terminal earns. See `splash-fit.render.test.tsx`.
+    expect(text).toContain(":::");
     expect(text).toContain("/help");
   });
 
--- a/src/tui/components/chat-log.tsx
+++ b/src/tui/components/chat-log.tsx
@@ -1,6 +1,7 @@
 import { Box, Text, measureElement, type DOMElement } from "ink";
 import { useEffect, useRef, useState, type ReactElement } from "react";
 import { useTerminalSize } from "../hooks/use-terminal-size.js";
+import { computeChatViewportRows } from "../layout.js";
 import type { TuiAction } from "../tui-action.js";
 import type { ChatMessage, TuiState } from "../tui-state.js";
 import { theme } from "../theme/theme.js";
@@ -15,15 +16,6 @@ import { SystemBubble } from "./system-bubble.js";
 import { ThinkingIndicator } from "./thinking-indicator.js";
 import { ToolCard } from "./tool-card.js";
 import { UserBubble } from "./user-bubble.js";
-
-/**
- * Rows of "chrome" outside the chat surface: status bar + prompt
- * meta-row + prompt input + prompt tail-cap + hotkey hint + a small
- * safety pad. Used to convert `terminal.rows` into the chat-area
- * viewport height. Slightly conservative — better to leave one empty
- * row than to clip the prompt.
- */
-const CHROME_ROWS = 8;
 
 interface ChatLogProps {
   state: TuiState;
@@ -89,7 +81,10 @@ export function ChatLog({ state, dispatch }: ChatLogProps): ReactElement {
   // All hooks must run unconditionally — only the JSX branches on
   // `isEmpty`. Compute viewport / measured-K / clamp regardless,
   // even when the early return for the splash branch fires below.
-  const viewport = Math.max(5, terminalSize.rows - CHROME_ROWS);
+  const viewport = computeChatViewportRows(
+    terminalSize.rows,
+    terminalSize.columns,
+  );
   // First-frame fallback for `K` until the post-mount `measureElement`
   // call returns the truth. Estimates are unreliable (text wraps, Yoga
   // collapses some margins, reasoning blocks expand mid-turn) so we
--- a/src/tui/components/logo-fit.test.ts
+++ b/src/tui/components/logo-fit.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it } from "vitest";
+import { LOGO_ART, TAGLINE, WORDMARK_ROWS } from "./logo.js";
+import { LOGO_METRICS, WORDMARK_WIDTH, type LogoVariant } from "./splash-fit.js";
+
+function measure(rows: readonly string[]): { width: number; height: number } {
+  return {
+    width: rows.reduce((acc, row) => Math.max(acc, row.length), 0),
+    height: rows.length,
+  };
+}
+
+/**
+ * `splash-fit.ts` picks a mark from numbers it keeps in `LOGO_METRICS`;
+ * the artwork itself lives in `logo.tsx`. If the two ever drift the
+ * breakpoints silently start lying, so measure the real rows here.
+ */
+describe("logo artwork", () => {
+  const variants: readonly LogoVariant[] = ["full", "small", "mini"];
+
+  it.each(variants)("matches the declared metrics for %s", (variant) => {
+    expect(measure(LOGO_ART[variant])).toEqual(LOGO_METRICS[variant]);
+  });
+
+  it("orders the variants strictly smallest-last", () => {
+    expect(LOGO_METRICS.full.width).toBeGreaterThan(LOGO_METRICS.small.width);
+    expect(LOGO_METRICS.small.width).toBeGreaterThan(LOGO_METRICS.mini.width);
+    expect(LOGO_METRICS.full.height).toBeGreaterThan(LOGO_METRICS.small.height);
+    expect(LOGO_METRICS.small.height).toBeGreaterThan(LOGO_METRICS.mini.height);
+  });
+
+  it("matches the declared wordmark width and keeps the tagline narrower", () => {
+    expect(measure(WORDMARK_ROWS).width).toBe(WORDMARK_WIDTH);
+    expect(TAGLINE.length).toBeLessThanOrEqual(WORDMARK_WIDTH);
+  });
+});
--- a/src/tui/components/logo-raster.test.ts
+++ b/src/tui/components/logo-raster.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it } from "vitest";
+
+import { LOGO_ART } from "./logo.js";
+import { rasteriseMark, toInkMask } from "./logo-raster.js";
+import { LOGO_METRICS } from "./splash-fit.js";
+
+const source = toInkMask(LOGO_ART.full);
+
+/** Ink coverage as a fraction of the box — a crude "does it look like the mark". */
+function density(rows: readonly string[]): number {
+  const total = rows.reduce((acc, row) => acc + row.length, 0);
+  if (total === 0) return 0;
+  const ink = rows.reduce(
+    (acc, row) => acc + [...row].filter((ch) => ch !== " ").length,
+    0,
+  );
+  return ink / total;
+}
+
+describe("rasteriseMark", () => {
+  it("returns exactly the box it was asked for", () => {
+    for (const [columns, rows] of [
+      [20, 12],
+      [13, 8],
+      [7, 4],
+    ] as const) {
+      const art = rasteriseMark(source, { columns, rows });
+      expect(art).toHaveLength(rows);
+      for (const line of art) expect(line).toHaveLength(columns);
+    }
+  });
+
+  it("keeps the mark's aspect ratio instead of stretching it", () => {
+    // The source is 34 cells wide and 20 tall = 34x40 half-block pixels.
+    // Asked for a box twice as wide as that shape needs, the drawing must
+    // stay its own shape and sit centred, not stretch to the edges.
+    const art = rasteriseMark(source, { columns: 60, rows: 12 });
+    const drawn = art.filter((row) => row.trim().length > 0);
+    const leading = Math.min(
+      ...drawn.map((row) => row.length - row.trimStart().length),
+    );
+    const trailing = Math.min(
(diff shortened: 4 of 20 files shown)


sosidudku1:
The arithmetic is well-floored — I traced 1×1, 20×5, 0×0 and the 100-column boundary and nothing negative or zero reaches Ink. No new resize listener, and the fit is computed from props rather than measured during layout, so there's no re-render loop.

**`layout.ts:118` — the sidebar row budget overflows short terminals, which is the symptom this PR exists to prevent.**

```ts
const usable = Math.max(2, rows - SIDEBAR_CHROME_ROWS);
```

The floor of 2 holds regardless of height, and both clamps below it floor at 1. Meanwhile `isSidebarVisible` gates on width only — height is never consulted.

On a 100×8 terminal (a split tmux pane, or a terminal docked under an editor) that yields `sessions: 2, tasks: 1`, and the rendered rail comes to roughly 10 rows — border, two headers, three list rows, two "↓ N more" footers and a margin — against 8 available. At 100×5 it's 10 against 5. Per the comment in this PR, Ink 7 overlaps rather than clips, so the frame garbles.

`SIDEBAR_CHROME_ROWS = 7` also under-counts the unconditional chrome plus the two optional footers. Gating visibility on height as well as width would be the direct fix.

plombeer31:
Fixed in d4be390.

**`isSidebarVisible(columns, rows)`** (`src/tui/layout.ts`) — now gates on height as well as width. `SIDEBAR_MIN_ROWS` = reserved chrome + one row per pane = 9. Below that the rail is not drawn at all. Both call sites updated: `computeChatWidth` (which also took a `rows` argument, so the splash and the rail keep agreeing about the boundary) and `tui-app.tsx`.

**`computeSidebarRowBudget`** — the `Math.max(2, …)` floor and both `1` floors are gone. `sessions` is capped at `usable - 1` so the Tasks pane keeps its last row, and `tasks` takes what is left; `sessions + tasks <= usable` now holds for every input.

**`SIDEBAR_CHROME_ROWS`** — I rendered the component instead of counting by eye, and the number is 5, not 7: two section headers, the blank row between the panes, and a "↓ N more" footer per pane. The left border costs columns, not rows — `sidebar.tsx` sets `borderTop={false} borderBottom={false}`. The old 7 was 5 + the status bar + one row of slack, which is where the other two rows in your count went; it was correct as a total, so the reserved rows are still 7, but it is now `SIDEBAR_CHROME_ROWS + SIDEBAR_OUTER_ROWS` with each half named and testable rather than one number with a prose derivation. Tall-terminal budgets are unchanged (16 → 6/3, 24 and up → 10/5).

Tests:

- `layout.test.ts` — the rail collapses one row below the threshold; hidden at 100×8, 100×5, 200×4, 1×1, 0×0; `computeChatWidth(120, 8)` returns the full 118; and for every height 0–60 where the rail is visible, `sessions + tasks + SIDEBAR_CHROME_ROWS + 1 <= rows`. The 100-column boundary tests are untouched apart from the new argument.
- `sidebar-fit.test.tsx` (new, sibling of `splash-fit.render.test.tsx`) — renders the real `Sidebar` at the computed budget for every height from 9 to 24 and asserts the frame is exactly `sessions + tasks + SIDEBAR_CHROME_ROWS` rows and never more than `rows - 1`. This is what pins `SIDEBAR_CHROME_ROWS` to the component rather than to a comment.

Checked under a PTY at the real sizes (`pyte`, `ATOMIC_AGENT_TUI_SKIP_LLAMA_SETUP=1`). Before, at 100×5, the rail's headers were overwritten and its orphaned rows interleaved with the prompt:

```
|  atomic-agent v0.2.2  |  ▸ Run  ·    Observe  ·    Manage|
|                                                          │ (no sessions yet)|
|  │                                                       │|
|  │ ❯  Type a message or `/` for commands…                │|
|  │                                                       │ (no active tasks)|
```

After, at both 100×8 and 100×5, the rail is gone and the frame is intact:

```
|  atomic-agent v0.2.2  |  ▸ Run  ·    Observe  ·    Manage|
||
|  │|
|  │ ❯  Type a message or `/` for commands…|
|  │|
```

`npm run lint` clean. Full suite: 4 failed files / 7 failed tests, all of them failing on the branch before this commit too (`send-message-concurrency`, `fs-glob-real`, `persist-embedding-hybrid-recall`, `tui-app` fixtures); the baseline run on the same machine was 9 files / 12 tests, the extra ones being flaky `git.*` temp-repo tests.

PR #152, opened 2026-08-19
fix(tui): reject an empty API key on the provider wizard key screen

plombeer31:
## Problem

The provider wizard accepts an empty API key.

On the `api_key` screen, Enter with an empty buffer just advances (`providers-wizard-key-bindings.ts`). The operator picks a model, presses Enter to save, and only then meets `throw new Error("API key is empty …")` from `saveProviderWizardToConfig` — three screens after the mistake.

That save-time check is also weaker than it looks. It passes whenever *any* key resolves from the environment, and the environment probe behind it was built from a fixed `openai-compatible` entry regardless of the service being configured:

```ts
resolveLlmProviderApiKey({ id: "openai-compatible", kind: "openai-compatible", … })
```

So an unrelated `OPENAI_API_KEY` satisfied the OpenRouter, AI/ML API and Gemini screens — and the same wrong key was handed to the model-list fetch.

## Changes

- New `src/tui/providers/providers-wizard-target.ts`: what the wizard state says about the service being configured — base URL, resolved key, env-var hint, whether a key is optional, and the key-screen error. `providers-wizard-key-bindings.ts` was sitting exactly on the 300-line limit, so the helpers moved rather than grew.
- Resolve the key from the **selected kind/preset**, so each service reads its own variable (`OPENROUTER_API_KEY`, `AIMLAPI_API_KEY`, `GEMINI_API_KEY`, or the preset's).
- Enter on the key screen refuses an empty or whitespace-only key and names the variable that would satisfy it instead. The next keystroke clears the message (the existing typing/backspace branches already reset `error`).
- Keyless services are untouched: local presets (LM Studio, Ollama) and keyless-listing ones (Nous, Ollama Cloud) skip or pass the screen exactly as before, and an empty screen is still fine when the service's own key is already in `.env`.
- `saveProviderWizardToConfig` keeps its throw as a backstop for callers that never see the screen.

## Tests

- `src/tui/providers/providers-wizard-target.test.ts` — empty/whitespace refused per kind and per preset; keyless presets exempt; `.env` key accepted; **an empty OpenRouter screen stays refused while only `OPENAI_API_KEY` is set**.
- `providers-wizard-key-bindings.test.ts` — Enter on an empty key holds the screen and sets the error, the next keystroke clears it, LM Studio still skips the screen entirely. Two existing walkthroughs now type a key before Enter, which is the behaviour change.

`npm run lint` clean; `npx vitest run src/tui/providers` 128 passed.


sosidudku1:
The extraction into `providers-wizard-target.ts` is clean, and scoping key lookup to the wizard's real kind is a genuine fix — a stray `OPENAI_API_KEY` answering for OpenRouter was a real trap.

One blocker before merge: **the new gate is stricter than the save-time check it fronts, and it breaks reconfiguring an existing provider.**

`apiKeyPhaseError` resolves the key through `keyLookupEntryForWizard`, which builds `{ id, kind, apiKeyEnvVar }` and never sets `apiKey`. So `resolveLlmProviderApiKey` skips its `entry.apiKey` branch entirely — a key that lives in `config.json` (which is where `upsertLlmProvider` puts it when the operator typed one, with nothing written to `.env`) is invisible to the gate.

Repro: add a provider by typing a key, then select it in the providers panel and press `c`. The wizard opens at `api_key` with an empty buffer; Enter now returns `API key required — …` and the phase stays put. On `main` the same input advances to `pick_chat_model`. Esc is not a way out — it rebuilds a fresh `pick_kind` state and drops `kind`/`baseUrl` — so there is no path to reconfigure that provider's model without retyping a key that is already saved.

`saveProviderWizardToConfig` already handles this case deliberately:

```ts
} else if (existing?.apiKey) {
  entry = { ...entry, apiKey: existing.apiKey };
}
```

Suggested fix: have the gate consult the existing config entry the same way the save path does — pass `existing` into the lookup, or check `existing?.apiKey || resolveLlmProviderApiKey(existing)` before refusing.

Worth adding a `configure`-mode test too: every new test here constructs `createProvidersWizardState("add", …)`, which is why this slipped through.

plombeer31:
Fixed in 3da0194.

**The gate.** `apiKeyForWizard` (`src/tui/providers/providers-wizard-target.ts`) now looks up the `config.json` entry a `configure` run would overwrite and resolves through it first — `storedEntryForWizard` finds it by `wizard.providerId`, then `resolveLlmProviderApiKey(stored)`, which covers both `existing.apiKey` and the entry's own `apiKeyEnvVar`. That is the union `saveProviderWizardToConfig` accepts, so `apiKeyPhaseError` can no longer refuse what the save would take. The lookup is skipped unless `mode === "configure"` and `providerId` is set, so the `add` gate and the kind-scoped probe are unchanged.

**Esc.** A configure run opens on `api_key`, so stepping "back" from it built a `pick_kind` screen the run never showed. `isWizardFirstScreen` (`providers-wizard-phases.ts`) names the screen a run opened at, and Esc there closes the wizard instead — same as Esc on the provider list when adding. Esc from any later phase still steps back as before.

**Hint text.** The key screen said "Leave empty only if the key is already in .env" to an operator whose key is in `config.json`. `emptyKeyMeaningForWizard` now says "Leave empty to keep the key already saved." for that case; the keyless-service and `.env` wordings are unchanged. Moved out of `providers-wizard.tsx` (369 lines) so it is testable.

**Tests.** `providers-wizard-target.test.ts` gets a `configure`-mode block on a temp state dir seeded through `upsertLlmProvider`: an empty buffer passes with the key in `config.json`; it passes when the key is under the entry's own `apiKeyEnvVar` and no preset supplies it; it still refuses with no key anywhere and for an id not stored yet; and a stored key does **not** satisfy an `add` run. `providers-wizard-key-bindings.test.ts` gets the flow version: Enter advances to `pick_chat_model` with the key saved, stays put with the error when it is not, Esc on the key screen closes, Esc past it still steps back. The three positive ones fail against the previous `apiKeyForWizard`.

`npm run lint` clean. `npx vitest run src/tui/providers/` 10 files / 140 tests pass. `src/tui/` 5 failures, the same set as `main` (splash-banner, chat-log, tui-app x2, persist-embedding-hybrid-recall). Full suite 17 failures, all pre-existing: that set plus `fs-glob-real`/`send-message-concurrency` (identical on the base commit) and the `os.git.*` tests, which pass in isolation on both.

PR #155, opened 2026-08-19
fix(tui): Esc aborts a running turn again

plombeer31:
## Symptom

<kbd>Esc</kbd> does not abort a running turn. The hint strip advertises `[esc] abort` for the whole time a turn is in flight, but the keypress is inert — <kbd>Ctrl</kbd>+<kbd>C</kbd> is the only way to stop a run.

## Cause

The abort lives in the chat editor's `onEscape`:

```ts
if (canAcceptMessage(state)) { …quit… } else { callbacks.onAbort(); … }
```

That `else` is unreachable. The editor is rendered with `disabled={!canAcceptMessage(state)}` — true for every status except `idle` — and `disabled` switches its `useInput` off (`isActive: focus && !disabled`). For the entire duration of a run the editor is deaf, so the only branch that could abort never executes.

Esc from a **pending approval** does abort correctly; that path lives in `handleApprovalKey` and is unaffected.

## Fix

Esc-to-abort moves into `handleAppKey`, which is subscribed independently of editor focus and disabled state.

Guards kept deliberately narrow:

- a pending approval returns earlier and keeps its own abort-with-deny semantics;
- overlays that own Esc themselves (slash palette, theme picker, session picker) keep it;
- precedence matches the hint strip, which resolves `running` before `uiMode === "debug"` — a turn in flight aborts rather than navigating back to Run.

## Evidence

Counting `onAbort` calls while driving the real `TuiApp` with a turn in flight:

| scenario | before | after |
|---|---|---|
| Esc on the Run screen, running | `abort=0` | `abort=1` |
| Esc on a debug tab, running | `abort=0` | `abort=1` |
| Esc while idle | `abort=0` | `abort=0` (unchanged) |

## Tests

`src/tui/escape-abort-running.test.tsx` — abort from chat, abort from a debug tab, and an idle no-op guard. Verified red before the change (2 of 3 failing) and green after. The existing 25 `app-key-bindings` tests still pass.

`npx tsc --noEmit` clean; no new failures in `npx vitest run src/tui` against `main @ 667dae1`.


plombeer31:
Cross-ref for whoever merges these: #156 ("keep the editor live during a turn") swaps the editor's `disabled={!canAcceptMessage(state)}` for `disabled={!canTypeMessage(state)}`, which is the exact mechanism this PR routes around — today the editor is deaf for the whole run, so its `onEscape` abort branch can never fire.

The two still compose (the abort is claimed in `handleAppKey`, which runs before the editor and returns early), but with #156 in place the editor's own `useInput` is a *live second subscription* during a run, so a single Esc would reach both layers and call `onAbort()` twice. It is idempotent, so nothing breaks — but whichever of these two lands second should delete the now-redundant abort branch in `onEscape` rather than leave both.

sosidudku1:
The stated cause checks out, which is worth confirming since it's the whole basis for the change: `PromptShell` gets `disabled={!canAcceptMessage(state)}`, and `multi-line-editor.tsx` gates its hook with `{ isActive: focus && !disabled }` plus an early `if (disabled) return;` — so while a turn ran, the editor's Esc handler was genuinely unreachable and the advertised `[esc] abort` was dead. Moving it to the global layer is the right fix.

**One regression: `app-key-bindings.ts:133` drops the scroll-reset precedence.**

The new branch matches on `status === "running"` with no `chatScrollOffset` check, and it now runs before the handler it supersedes — which checked scroll first:

```ts
if (state.chatScrollOffset > 0) {
  dispatch({ type: "chat_scroll_reset" });
  return;
}
```

whose comment states the intent: "Esc with the chat scrolled away from the bottom snaps back to the latest reply before doing anything else — avoids a confused 'why didn't my Esc abort?'".

So: submit a message, PageUp to read back through the streaming answer (allowed while running), press Esc expecting the documented snap-back — and the in-flight turn is destroyed instead. Measured with all three Esc PRs applied, the abort callback fired once in that sequence; on unmodified `main`, zero.

Adding `state.chatScrollOffset === 0` to the new condition, or resetting scroll and returning when the offset is non-zero, restores it.

*(Reviewed alongside #153 and #154: no textual or semantic conflict between the three — they cherry-pick onto `main` cleanly and their tests pass together, and Esc can still quit the agent from idle chat.)*

plombeer31:
Fixed in a117dc3.

**Scroll-reset precedence.** The scroll check now lives inside the running-abort branch, so the rung order the editor documented survives:

```ts
if (key.escape && state.status === "running" && !overlays) {
  if (state.uiMode === "chat" && state.chatScrollOffset > 0) {
    dispatch({ type: "chat_scroll_reset" });
    return true;
  }
  callbacks.onAbort();
  ...
}
```

Doing the reset here rather than just adding `chatScrollOffset === 0` to the condition: with the guard alone the key would fall through to a `disabled` editor, so submit → PageUp → Esc would do nothing at all and the second Esc would still not abort. The `uiMode === "chat"` clause is because nothing resets `chatScrollOffset` on a mode switch — on a debug tab the chat is off-screen, and snapping an invisible offset back would make Esc look dead there instead of aborting.

**Duplicate abort branch.** Removed from the editor's `onEscape` per the cross-reference note, since the claim now lives in `handleAppKey`. The editor keeps its other Esc duties (theme picker / session picker / slash palette close, pending-approval bail, scroll reset, idle quit). Today the branch is unreachable anyway; with #156 it would have been a live second subscription firing `onAbort` twice per keypress.

**Tests.** `src/tui/app-key-bindings.test.ts` gets three unit cases (running + offset 0 → abort once; running + offset 8 → `chat_scroll_reset`, no `abort_requested`; running + debug tab + stale offset → abort). `src/tui/escape-abort-running.test.tsx` gets the reported sequence end-to-end through Ink: submit, 40 lines of chat, PageUp, Esc → abort fires 0 times, second Esc → 1. The existing chat-surface case tightened from `toBeGreaterThan(0)` to `toBe(1)` to pin the no-double-abort contract.

Not vacuous — Esc is written to a real stdin and every assertion sits behind a 60 ms settle past Ink's 20 ms lone-Esc flush. Reverting the source hunk fails both new scroll tests; applying the guard-only variant (`chatScrollOffset === 0` with no reset) fails three.

`npm run lint` clean. `vitest run src/tui/app-key-bindings.test.ts src/tui/escape-abort-running.test.tsx` → 32/32. Full `src/tui` sweep: 965 passed, 5 failed — all five (splash-banner, chat-log splash, two tui-app smoke fixtures, `localModels.embeddings`) reproduce identically at 21b77ff with the diff stashed.

PR #156, opened 2026-08-19
fix(tui): keep the editor live during a turn and queue what you type

plombeer31:
## The problem

Press Enter and the input goes dead until the turn finishes. Not "your
message is queued" — the editor rejects every keystroke, so you cannot
even draft the next message while the agent works.

## Why it was dead

`canAcceptMessage(state)` is just `status === "idle"`
(`src/tui/tui-state.ts`). It gated two things at once:

- `submit-handler.ts` — `if (!canAcceptMessage(state)) return;` dropped the submit on the floor.
- `tui-app.tsx` — `disabled={!canAcceptMessage(state)}` on `PromptShell`, which becomes `isActive: focus && !disabled` on the Ink `useInput` subscription. That is what eats the keystrokes.

The queue this needs **already existed and was unreachable**:
`ChatOrchestrator.sendMessage` pushes into `this.queue` whenever
`currentController` is set, and `runOneTurn` drains it FIFO. Nothing
ever got that far, because the UI rejected the submit first.

## What changed

- **Split the selector.** `canAcceptMessage` keeps its meaning ("a new turn may start now") and drives the submit branch. New `canTypeMessage` (`status !== "quitting"`) drives the editor. Typing is live for the whole run.
- **A submit made mid-run dispatches `message_queued`, not `message_submitted`.** This distinction is load-bearing: `message_submitted` calls `startNewRun`, which wipes `feed`, `reasoning`, `streamingToolCards` and `streamingAssistantText`. Reusing it would blank the screen of the turn you are reading. There is a regression test for exactly this.
- **The orchestrator is the source of truth for the queue.** It re-publishes on every push, drain, clear, session switch and quit via `queue_changed`, so the UI mirrors what will actually run rather than keeping an optimistic copy that can drift.
- **You can see it.** Parked messages render as a dim strip above the prompt (`QueuedMessages`), the hint strip shows what Enter does mid-run and how many are parked, `/queue` lists them, `/queue clear` drops them.

Runtime, agent loop and prompt are untouched — this is a TUI-layer fix.

## Verification

- `npm run lint` clean.
- 59 tests across the six touched suites pass, including a
  `ChatOrchestrator` test that pins drain order and one that pins the
  live turn's `feed` surviving a queued submit.
- Full suite: no new failures against `main` (see the note in the PR
  thread — `main` itself is flaky at ~10 failures/run).


sosidudku1:
The queueing wiring is sound — the orchestrator stays the single source of truth, ordering is exactly-once and in-order, and the editor change leaks no key handler or raw-mode state since it only flips an existing `disabled` prop.

One blocker: **aborting a turn drains the queue instead of discarding it, so Esc doesn't stop the agent.**

`abortCurrentTurn()` is only `this.currentController?.abort()`. The rejection is swallowed by the catch in `runOneTurn`, and execution falls straight through to `this.queue.shift()` and launches the next parked message.

Concretely: type "run the migration", queue two more messages while it streams, then hit Esc because the migration is wrong — the next queued message starts immediately. Stopping requires one Esc per parked message. Confirmed with a probe test at this head: `runTurn` was called with `["running", "parked-a"]` after a single abort.

`quit()` gets this right four lines below — it sets `quitting` and clears the queue before aborting — so this is just making the two paths agree.

This is newly operator-visible precisely because this PR is what makes the queue reachable; before it, the disabled editor meant nobody could fill it.

**Worth considering:** the queue is unbounded and `emitQueue()` copies the whole array on every push, so N parked messages is O(N²) work, with every message replayed as a real turn after the drain. The disabled editor used to be a de-facto cap and this PR removes it — a bound with a visible "dropped N" notice would settle it.

plombeer31:
Fixed in e21b2d0.

**Abort.** `abortCurrentTurn()` now clears the queue and re-publishes it before calling `abort()` — the same order `quit()` uses four lines below. I reproduced your probe first: at `1fe88d3`, one abort with two parked messages left `runTurn` called with `["running", "parked-a"]`. It is now `["running"]`, pinned by `chat-orchestrator.test.ts > ChatOrchestrator abort > discards parked messages instead of draining them into the next turn`. Reverting just that hunk fails only that test.

The drop is announced rather than silent: `aborted: dropped 2 parked messages`, as a `runtime_info` line plus a warn `system_message` — the pair `runOneTurn`'s catch already uses, because in chat mode the feed is not on screen. Nothing is emitted when the queue was already empty.

**Bound.** `MAX_QUEUED_MESSAGES = 20`. Past the cap a submission is refused with `queue: full at 20 — dropped N message(s); Esc stops the run, /queue clear empties it`, where N is a running count that resets on the first push that fits, so a paste storm reads as one escalating counter instead of 30 identical lines. 20 because every parked message is replayed as a real turn, which makes the cap really "how long an unattended drain may be" — and 20 is well past anything a human types while watching one turn stream.

The refusal still emits `queue_changed` even though the queue did not change: the reducer inserts optimistically on `message_queued`, so without the re-publish the strip would keep showing a message that is never going to run. Pinned separately.

**The O(N²) copy.** Left it, deliberately. The bound caps it at 20 elements per push (~400 element copies to fill a queue), which retires the complexity concern. Removing it means `queue_changed` carries a push/shift/clear delta and the reducer applies it — queue arithmetic back in the reducer, i.e. exactly the drift this PR removed by making the orchestrator authoritative. Reasoning is in the comment on `emitQueue`.

`npm run lint` clean. `chat-orchestrator` + `reduce-ui-actions` + `submit-handler`: 32 passed / 0 failed, 5 of them new. Full `src/tui`: 987 passed, 6 failed — all reproduce at stock `1fe88d3` (splash-banner and chat-log fixtures, two tui-app smoke fixtures, `localModels.embeddings`) plus one ink-timing tui-app test that only fails under the parallel full-suite run and passes in isolation on stock HEAD. Whole suite: 4088 passed, 14 failed, all in the pre-existing buckets (the above plus `src/tools/os/git/*`, which times out in this sandbox).

PR #157, opened 2026-08-19
feat(runtime): mid-turn steering — reach the turn that is already running

plombeer31:
## The problem

Per-session FIFO is the right answer for *starting* turns and the wrong
one for *correcting* one. Today a message sent while the agent is
working cannot reach the model at all until the turn closes, so an
operator watching it head the wrong way has exactly one lever: abort and
start over.

## The seam

`AgentLoop.runTurn` already carries `pendingNotice` → `transientNotice`
→ `### notice` in the prompt's **variable tail**
(`build-prompt.ts`). The no-progress loop detector already writes there.
It is in the tail, so writing to it costs **no KV-cache invalidation**
(architectural invariant 2).

## What changed

`SteeringInbox` (`src/runtime/steering-inbox.ts`) — an out-of-band,
per-session mailbox. Deliberately **not** a second queue; there is still
exactly one path into `AgentLoop.runTurn`.

- **It never starts a turn.** `runtime.steer(sessionId, text)` returns `false` and queues nothing when the session has no turn in flight. "Not steered" is a signal the caller acts on, never a silent drop.
- **It lands at a step boundary.** The loop drains it at the top of every step, before building that step's prompt — never mid-inference, never mid-tool-call. A turn parked in a long `os.shell.run` will not react until that call returns. That is inherent to one-inference-per-step, not a bug.
- **It writes to the transcript.** Each drained message becomes a real `user` `ConversationTurn`. The transcript must not lie about what the operator said. `packConversation` already pins the last user turn visible; `findCurrentMacroTurnStart` folds it into the macro-turn in progress. **It does count toward reflection segmentation** — that is the intended reading, it is a real user message.
- **The text is repeated in `### notice`, composed with (not over) the loop-detector's text.** The duplication is deliberate: `### notice` is the last block before `### respond`, which is the one place a 30B local model reliably acts on. Long pastes are clipped inline and point back at the transcript copy.
- **Nothing is lost.** A message pushed after the loop's final drain — during the last inference, or into a turn cancelled before it stepped — comes back on `RunTurnResult.undelivered` for the caller to re-route. `shutdown()` clears the inbox so a stale steer cannot resurface in a later process.
- **Bounded.** 16 pending per session; `push` refuses past the cap rather than evicting the oldest, so the caller learns the message did not land.

No producer is wired up in this PR — the TUI gesture and the
sidecar/HTTP endpoints land separately. Without the `steeringInbox` dep
the loop is byte-identical to before, pinned by a test.

## Verification

- `npm run lint` clean.
- 218 tests across `src/agent/` + `src/runtime/` pass.
- New: 9 inbox tests, 8 notice-formatting tests, 8 loop-integration tests (injection at the next step, one-shot notice, transcript turn, undelivered on reply / on cancel, no-op without the dep), 3 runtime-level tests.
- `AGENTS.md` gains a §"Mid-turn steering" and an ownership-table row.


sosidudku1:
The core mechanism is right: injection happens at a step boundary rather than mid-inference, the steer text lands in the variable tail so the stable prefix and KV-cache are untouched, and the inbox is bounded at 16 and refuses rather than evicting.

**`bootstrap.ts:2106` — `steer()` has a lost-update race with the end of a turn.**

```ts
const steer = (sessionId: string, text: string): boolean => {
  if (!turnController.isBusy(sessionId)) return false;
  return steeringInbox.push(sessionId, text);
};
```

Check-then-act, and the two facts move at different times: `flushSteering` runs inside `runTurn`, while `busy.delete` runs later in `TurnController`'s `finally`. In that window `isBusy` is still true, so `steer` returns `true` and the operator's UI shows the message accepted — but the final drain already happened.

The message then sits in the inbox until the *next* turn on that session, where it arrives at step 0 with a notice telling the model the user sent it "while you were working", about a turn that ended long ago. It isn't in `undelivered` either, since it arrived after the flush.

This is the case AGENTS.md describes as "Nothing is silently lost."

**Two smaller ones:**

- AGENTS.md says callers must re-route `undelivered` and that "the TUI pushes it onto its pending-message queue" — but this PR touches no file under `src/tui/`, `src/cli/`, `src/http/` or `src/sidecar/`, and `grep` shows `undelivered` read only by tests. As merged, every undelivered steer is dropped.
- The new `steer_applied` event has no consumer; `reduceAgentEvent` ends in `default: return state`, so the inline rendering its doc comment promises doesn't happen, and the non-exhaustive default means TypeScript doesn't flag it.

plombeer31:
All three addressed. Head is `a520d9f`.

**1. The race (`bootstrap.ts` `steer()`) — `725307c`**

Approach taken: **the inbox owns a per-session acceptance window, and the same indivisible call closes it and performs the final drain.** `AgentLoop.runTurn` calls `steeringInbox.open(sessionId)` on entry; every exit path goes through `flushSteering`, which is now `closeAndDrain(sessionId)`. `push` refuses whenever the window is shut, so `steer()` is a single atomic decision — it no longer consults `turnController.isBusy` at all. That was the stale half of the pair: `busy` is cleared in the controller's `finally`, after `run()` settles, while the loop's last drain happens earlier inside `runTurn`. Nothing can now land "after the drain but while busy" and be told yes.

A `return` expression is evaluated before the enclosing `finally`, so the normal paths still report exactly what was pending as `undelivered`. The wrapper's `finally` covers the throw path, where there is no result to carry `undelivered`: it closes the window and warn-logs anything stranded, rather than leaving the session open for later steers nobody would drain.

Rejected alternative: **a per-session turn generation captured before the push.** To be correct the generation still has to be validated *inside* `push` — comparing it in the caller is the same check-then-act — so it collapses to the same mechanism, with a counter to reason about instead of a boolean and a wider API. The boolean window is the same guarantee with less surface.

Test: `bootstrap.test.ts` › "refuses a steer that lands after the turn's final drain". `runTurn` is `enqueue({ run: () => executeTurn(...) })`, so spelling that composition out by hand puts an assertion *between* the loop's final drain and `busy.delete` — real `TurnController`, real `SteeringInbox`, real `AgentLoop`, real `runtime.steer`, no sleeps and no timing luck. It asserts `isBusy` is still true at that point, so the window is genuinely open. The same test also pins the working direction: a steer sent from inside step 0's inference is accepted and lands in step 1.

Proof it is not vacuous: delete the one line `if (!this.accepting.has(sessionId)) return false;` from `push` and it fails with `expected [ true ] to deeply equal [ false ]` — the old behaviour. Relax that first assertion and it fails at the next one with `steer_applied { text: "too late, stop", stepIndex: 0 }` on the *following* turn: exactly the symptom described in the review.

**2. `undelivered` had no consumer — `bcc46dd`**

`ChatOrchestrator` now holds both ends. `sendMessage` offers a message typed during a turn to that turn first (`runtime.steer`) and falls back to its own pending queue when that returns `false`; `rerouteUndelivered` puts anything handed back at the **head** of that queue — those were accepted as steers for the turn that just ran, and anything already queued was typed after `steer` had started refusing, i.e. later. Pinned by `src/tui/chat-orchestrator-steering.test.ts` (5 cases, both hunks verified by reverting each and watching 2 tests go red).

One thing to be explicit about: the keyboard gesture is unchanged. `canAcceptMessage` still accepts submissions only while idle, so today the TUI reaches this path programmatically. Opening the input mid-turn needs its own action — `message_submitted` currently resets the run view — and that is the separate TUI-gesture change. What lands here is the re-route contract AGENTS.md already promised, not the gesture. Left `src/http/` alone; that surface is #160.

**3. `steer_applied` had no consumer — `a520d9f`**

Added the case: a user bubble plus a feed line naming the step it was folded into, and none of the per-turn resets `user_message` implies (no `startNewRun`, no step-counter reset) — it is a correction to the turn in flight, not the start of one.

On the exhaustiveness question: I did make both switches in `agent-event-reducer.ts` end in `const unhandled: never = event`. Blast radius was small — it needed four events listed as explicit no-ops (`loop_detected`, `prompt_built`, `llm_completed`, `llm_raw_completion`) — and it returns `state` at runtime rather than throwing, because a UI reducer must not die on an event it does not recognise. It immediately found a second instance of the same class: `batch_trimmed` was also being swallowed. That one gets a feed line, since otherwise the dropped calls reappear one-by-one on later steps with no explanation for why the batch shrank.

AGENTS.md updated in all three places: `steer()` is no longer described as gated on `isBusy`, the re-route obligation names the code that now honours it, and the `steer_applied` rendering + exhaustiveness rule are written down.

**Verification** (node 25.7.0)

- `npm run lint` — clean.
- `npx vitest run src/runtime/ src/agent/agent-loop-steering.test.ts src/agent/steer-notice.test.ts src/tui/chat-orchestrator-steering.test.ts src/tui/agent-event-reducer.test.ts` — **11 files, 134 tests, all pass**.
- `npx vitest run src/agent/ src/http/ src/sidecar/ src/cli/ src/tasks/` — 34/35 files pass; the one failure is `sidecar/send-message-concurrency.test.ts`, which fails identically at stock `d67f090` (`expected [ 'm1', 'm2', 'm3' ] to deeply equal [ 'm1', 'm2' ]`).
- `npx vitest run src/tui/` — 99/103 files, 965/970 tests pass. The 4 failing files (`tui-app`, `splash-banner`, `chat-log`, `persist-embedding-hybrid-recall`) fail with the identical 5 assertions at stock `d67f090`.

sosidudku1:
Thanks — the race fix is the real thing, not a narrowing. Moving acceptance off `TurnController.isBusy` onto an `accepting` set that the loop itself opens, with `closeAndDrain` shutting the window and draining in one synchronous step, removes the interleaving entirely rather than making it less likely. The exhaustive `const unhandled: never = event` is also the right shape — it returns `state` instead of throwing, so an unknown event from a peer can't take the UI down.

One thing the new arrangement introduces, worth a look before merge.

**The window now opens later than the TUI considers a turn in flight**, so a steer sent in that gap is refused and demoted to the back of the queue.

`ChatOrchestrator.runOneTurn` sets `this.currentController = controller` and then awaits `runtime.runTurn`. `open()` is the first statement of `AgentLoop.runTurn`, but it is only reached after `turnController.enqueue` → `await waitOrAbort(previousSettled, signal)` → `executeTurn` → `ensureRecorder` → `turnContext.run`. That chain spans several microtask ticks, and if a previous turn on the session is still settling, `waitOrAbort` parks for the whole remainder of that turn.

A message submitted in that span sees `currentController` set, so it takes the steer branch — but `accepting` has no entry yet, `steer()` returns false, and it lands via `this.queue.push` *behind* anything already queued. Before this change `isBusy` was true across that span and the same message was accepted.

Nothing is lost, so this is not the old bug returning. But the ordering the operator intended is, and they don't get the "steering the running turn" feedback either. Opening the window where the orchestrator commits to the turn — or having `steer()` fall back to a front-of-queue insert when the window isn't open yet — would close it.

plombeer31:
Fixed in `ff03e24`. I took the second shape — front-of-queue insert when the window isn't open — and rejected opening the window at the commit point.

**Why not open earlier.** The window is a per-session single slot. Opening it before the submission owns the session lock aliases two turns onto one window, and the parking case you describe is exactly when that bites: the turn still running would drain a message meant for the parked one at its own step boundary, and its `closeAndDrain` would then carry off the parked turn's pending steers as *its* `undelivered`. Making that correct needs a per-turn window, which is the generation counter 725307c already considered and rejected. Separately it is unfixable on the abort-while-parked path: `run()` never executes, so nothing closes the window and there is no result to hand anything back on — a message `steer` said yes to would be dropped, which is the defect bcc46dd fixed. Front-insert keeps `open`/`closeAndDrain` paired inside `AgentLoop.runTurn`, exactly once, throw path included.

**What changed.** `ChatOrchestrator.sendMessage`'s fallback now splices into the front of the pending queue instead of pushing to the back — same rule and same reason as `rerouteUndelivered`. `steeredAhead` tracks how many leading entries were re-routed for the turn in flight, so two messages sent into the same gap keep their typing order rather than reversing, and it resets when a turn starts, which is what turns "aimed at the turn in flight" into ordinary backlog. `undelivered` still goes ahead of all of them: those were accepted while the window was open, i.e. earlier.

**Feedback.** The refusal now emits its own `steering the running turn — it cannot take this one, so it runs as the next turn`. `sendMessage` still reads exactly one fact — it does not ask `steeringInbox.isOpen` to tell "already shut" from "not yet open" from "inbox full" apart, since that is a second fact read at a different moment than the one `steer` acted on. Hence one line that is true for all three.

**Tests.** `src/tui/chat-orchestrator-steering.test.ts` gains a harness on the real `TurnController` and real `SteeringInbox`, with only the loop body stubbed in the shape `AgentLoop.runTurn` has (`open` on entry, `closeAndDrain` on the way out, then a settle phase for the session save and the controller's own `finally`). Gates, no sleeps: an out-of-band turn takes the lock, the TUI turn parks in `waitOrAbort` behind it, the occupant drains and closes, and the steer lands with the loop that will serve it still parked. Reverting the `chat-orchestrator.ts` hunk fails both new behaviour tests — no acknowledgement, and the correction runs after backlog left by an earlier turn. `npm run lint` clean; `src/runtime/ src/agent/agent-loop-steering.test.ts src/tui/chat-orchestrator-steering.test.ts` = 9 files / 103 tests, and the full suite matches stock HEAD exactly (6 files / 8 tests, the known-broken set).

**For #160.** This does not move the window, so the reconciliation there is unchanged in mechanism — but it does add a precedent worth applying. The caller's notion of "a turn is in flight" is *deliberately* wider than the window, so `turnController.isBusy(id)` is not just stale, it answers a different question: it can be true while `steer` correctly refuses (parked, or past the final drain) and it tells the route nothing about whether the message will be folded in. `POST /steer` should drop the pre-check, call `steer` and branch on its return — and treat `false` as "not folded into the running turn", not as "no turn running". Whatever it does with a refusal is the HTTP side's call; I have not touched `src/http/`.

PR #158, opened 2026-08-19
fix(tui): Esc in the chat editor clears the draft instead of quitting the agent

plombeer31:
> **Stacked on #153 ("Esc returns to Run from Observe tabs").** Both touch `onEscape`. Merge that one first; this branch is built on it.

## Symptom

A single <kbd>Esc</kbd> on the Run screen terminates the agent as soon as the session is idle — no confirmation, and no hint strip anywhere advertises it. <kbd>Ctrl</kbd>+<kbd>C</kbd> deliberately asks twice before it kills anything; Esc did it on the first press, and took any half-typed message with it.

It is reachable straight out of ordinary navigation, which is how I hit it:

```
/skill  →  Esc  →  back to Run   (correct)
        →  Esc  →  agent exits   (the natural "and out of here too")
```

Same from `/model`, and from every overlay: closing the slash palette, theme picker, session picker or sidebar focus with Esc leaves you one press away from an unannounced exit.

## Cause

`onEscape` ends in:

```ts
if (canAcceptMessage(state)) {
  callbacks.onQuit();
  dispatch({ type: "quit_requested" });
}
```

Esc means *cancel / back one level* on every other surface of this TUI. The one place it meant "exit" was the one place nothing said so.

## Fix

Esc on an idle Run screen clears the draft and no-ops on an empty buffer, reusing the same `input_changed` reset the submit handler dispatches — which also clears the input-history cursor.

Quitting stays on <kbd>Ctrl</kbd>+<kbd>C</kbd> twice and `/quit`. The abort path for a non-idle session is unchanged.

No hint-strip change: the chat row is already at its six-chip cap and still correctly shows `[ctrl+c] quit`.

## Evidence

Counting `onQuit` calls while driving the real `TuiApp`:

| scenario | before | after |
|---|---|---|
| Esc, idle Run screen | `quit=2`, app gone | `quit=0` |
| Esc with a draft typed | `quit=2`, draft lost | `quit=0`, draft cleared |
| `/skill` → Esc → Esc | 2nd press exits | stays on Run |
| Esc ×2 from slash palette / theme picker / session picker / sidebar | 2nd press exits | stays on Run |

## Tests

`src/tui/escape-chat-editor.test.tsx` — idle no-quit, draft-clearing, and the reported panel → Run → Esc → Esc sequence. Verified red before the change (3 of 3 failing) and green after.

`npx tsc --noEmit` clean; no new failures in `npx vitest run src/tui` against `main @ 667dae1`.

### DIFF
--- a/src/tui/components/hotkey-hint.test.tsx
+++ b/src/tui/components/hotkey-hint.test.tsx
@@ -1,3 +1,4 @@
+import { Box } from "ink";
 import { render } from "ink-testing-library";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import React from "react";
@@ -16,8 +17,24 @@ const MAC_SCROLL_KEY = "fn+↑↓";
 const OTHER_SCROLL_KEY = "pgup/pgdn";
 const SCROLL_KEY_PATTERN = /fn\+↑↓|pgup\/pgdn/;
 
-function renderHint(state: TuiState): string {
-  const { lastFrame, unmount } = render(<HotkeyHint state={state} />);
+/**
+ * Wide enough that no chip is shed — these cases assert *which* chips a
+ * state offers, not how the row degrades. Narrow behaviour has its own
+ * describe block below.
+ */
+const WIDE = 200;
+
+/**
+ * The strip is rendered inside a column-direction Box exactly as
+ * `TuiApp` renders it, so Ink resolves the same width and the frame we
+ * assert on is the frame the operator sees at `columns`.
+ */
+function renderHint(state: TuiState, columns: number = WIDE): string {
+  const { lastFrame, unmount } = render(
+    <Box width={columns} flexDirection="column">
+      <HotkeyHint state={state} width={columns} />
+    </Box>,
+  );
   const out = (lastFrame() ?? "").replace(ANSI, "");
   unmount();
   return out;
@@ -29,6 +46,15 @@ function chatState(overrides: Partial<TuiState> = {}): TuiState {
     uiMode: "chat" as const,
     ...overrides,
   };
+}
+
+/** Chips are `[key] label`; counting the brackets counts the chips. */
+function chipCount(frame: string): number {
+  return (frame.match(/\[/g) ?? []).length;
+}
+
+function widest(frame: string): number {
+  return Math.max(0, ...frame.split("\n").map((line) => line.length));
 }
 
 function fakeApproval(): ApprovalRequest {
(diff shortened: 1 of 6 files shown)


sosidudku1:
The change is correct and well-scoped: quitting stays reachable via double Ctrl+C and `/quit`, the earlier Esc branches are untouched, and the three new tests pass.

Two things worth deciding rather than leaving to branch order.

**Esc with a draft typed *and* a turn running.** Abort wins and the draft survives untouched; a second Esc, now idle, finally clears it. The branch reads `if (canAcceptMessage(state))` — run status — and never inspects `inputValue`. With #156 keeping the editor live during a turn, "draft present while running" becomes the normal state rather than a corner case, and nothing on screen disambiguates which meaning Esc has. Either order is defensible; it just shouldn't fall out of the `if` ordering by accident.

**The hint strip wasn't updated.** The PR's own rationale is that the old behavior "was a trap: no hint strip ever advertised it" — but the running strip still shows `esc / abort` with no mention that a draft survives, and the idle strip has no `esc` chip at all. So the new draft-clearing behavior is undiscoverable in exactly the way the PR is complaining about. An `esc / clear draft` chip on the idle strip when `inputValue` is non-empty would close it.

plombeer31:
Addressed in 7d198a8.

**Precedence — abort wins, draft survives.** While a turn is in flight Esc aborts and leaves `inputValue` alone; the next Esc, now idle, clears it. Abort is the destructive, time-critical action, and a draft is cheap to keep. The branch now reads `!canAcceptMessage(state)` first, bound to a `turnInFlight` local so the condition says "a turn is running" rather than "we can accept a message", with the rationale in a comment next to it. Pinned by `escape-chat-editor.test.tsx` → "aborts the turn and keeps the draft when both are on the table": running + draft + Esc calls `onAbort` once and the frame still contains the draft; a `turn_finished` and a second Esc clear it without a second abort.

**That precedence was unreachable, which is worth knowing.** The chat editor is `disabled` for the duration of a turn, so its Ink `useInput` unsubscribed — and on the chat surface it holds the only Esc subscription (`handlePanelEscape` bows out on `editorFocus`, `handleAppKey` has no chat-mode Esc branch). Esc during a turn did nothing at all, so `esc / abort` on the running strip was already a promise the TUI could not keep, on this branch and before it. `multi-line-editor.tsx` now forwards Esc while disabled and still swallows every key that could touch the buffer. Verified by reverting that hunk: the new precedence test fails with `expected +0 to be 1`.

**Strip, idle.** A non-empty buffer swaps `[/] commands` for `[esc] clear draft` rather than adding a seventh chip. The swapped chip is the right one to lose: with a non-empty buffer `/` no longer opens the palette (`slashPrefix` only fires on a leading slash), so it is inert in exactly the state that needs the new chip. Empty buffer keeps `[/] commands` and shows no clear-draft chip.

**Strip, running.** Esc has one meaning there, so the chip stays `abort` — but reads `abort, draft kept` when there is a draft, which is the part the screen could not tell you.

**Width.** Checking how the strip degraded turned up a bigger problem than the new chip: it never fitted. Ink wraps an over-wide row instead of clipping it, and the idle footer is 108 columns of chips — two rows at 100 columns, three garbled ones at 60 — while `debug-pane` budgets the strip as exactly one row (`APP_CHROME_ROWS`, with `RENDER_SAFETY_ROWS` already there to absorb the wrap). Chips now carry a `shed` rank and are dropped in that order until the row fits, and the row is `truncate-end` so the essential remainder clips instead of wrapping. `TuiApp` passes the chat column, not the terminal: `columns - root gutter - sidebar`.

What that looks like (mac spelling, chat column width):

```
idle, 118   [enter] send · [alt+enter] newline · [tab] sidebar · [fn+↑↓] scroll · [/] commands · [ctrl+c] quit
idle, 78    [enter] send · [alt+enter] newline · [/] commands · [ctrl+c] quit
idle, 58    [enter] send · [/] commands · [ctrl+c] quit
idle, 38    [enter] send  ·  [/] commands  ·  [ct…

draft, 118  … · [fn+↑↓] scroll · [esc] clear draft · [ctrl+c] quit
draft, 78   [enter] send · [alt+enter] newline · [esc] clear draft · [ctrl+c] quit
draft, 58   [enter] send · [esc] clear draft · [ctrl+c] quit

running+draft, 118  [fn+↑↓] scroll · [esc] abort, draft kept · [ctrl+c] abort
running+draft, 58   [esc] abort, draft kept · [ctrl+c] abort

debug, 118  [tab] next panel · [shift+tab] prev panel · [esc] back to Run · [/] commands · [ctrl+c] quit
debug, 78   [tab] next panel · [esc] back to Run · [/] commands · [ctrl+c] quit
```

Shed order is scroll first (the wheel already does it), then the sidebar hint (the same narrow terminals collapse the sidebar anyway, `SIDEBAR_MIN_COLUMNS`), then newline; `enter / send`, the esc chip and `ctrl+c / quit` are unshedable. Debug sheds `shift+tab` first because "prev panel" is guessable from "next panel".

**Verification.** `npm run lint` clean. Nine new tests, all non-vacuous — Ink's 20 ms lone-Esc deferral is waited out via the file's existing `settle()`, and each hunk was reverted to confirm the matching tests fail: editor Esc forwarding → precedence test fails; draft-chip logic → 4 strip tests fail; width fit + truncate → 3 narrow tests fail; making abort also clear the draft → the "draft survives" assertion fails. Full suite 4093 tests, same failing files as stock HEAD (splash-banner / chat-log / tui-app fixtures, `localModels.embeddings.url`, a dev-machine `fs-glob` path, `send-message-concurrency`, and the `os/git/*` sandbox timeouts); the extra `tui-app` "Esc from an idle Manage panel" failure reproduces at 2eebb34 in 2 of 4 runs.

PR #159, opened 2026-08-19
feat(tui): choose whether Enter steers the running turn or queues behind it

plombeer31:
> **Stacked.** This branch contains the commits from #156 and
> #157 as well, because it needs both. Merge those first; the commit
> that belongs to *this* PR is the last one
> (`feat(tui): choose whether Enter steers the running turn or queues behind it`).

## The problem

With the editor live for the whole turn (#156), Enter needs a
meaning while the agent is working — and there are two reasonable ones.
"Fold it into what you're doing now" and "run it next" are different
intentions and the operator is the only one who knows which they meant.

## The gesture

Default is **steer**: someone who types *while* the agent is working is
usually reacting to what they see it doing.

- **Ctrl+T** flips the mode in-app and persists it to `tui.whileBusySubmit`.
- The prompt meta-row shows the live mode (`⏎ steer` / `⏎ queue`) whenever a turn is running; the hint strip offers the flip.
- `/steer <msg>` and `/queue <msg>` land one message in the other mode without changing the default. Bare `/steer` / `/queue` switch it.

### Why a toggle instead of Alt+Enter

Alt+Enter was the obvious binding and it is not available:
`multi-line-editor.tsx` treats Return with **any** modifier
(`key.meta || key.shift || key.ctrl`) as "insert newline", so Alt+Enter,
Shift+Enter and Ctrl+Enter are all already taken. Binding one would cost
multi-line input. An explicit, visible toggle is the honest alternative.
Ctrl+T also had to join `isGlobalHotkey` so the editor does not swallow
it as literal text.

## Two things that keep a message from being lost or double-rendered

- `ChatOrchestrator.steerMessage` falls back to the queue whenever `runtime.steer` returns false — the turn can end between the keypress and the dispatch — and re-queues anything handed back on `RunTurnResult.undelivered`.
- The user bubble is rendered on the `steer_applied` event, **not** optimistically at submit time. So a steer that misses the turn and falls back to the queue renders exactly once, when it actually reaches the model.

## One drive-by

The hint strip stopped fitting on one terminal row at 80 columns once
the new chips were added, and a wrapped strip pushes the prompt down.
Labels are terse now, and an armed Ctrl+C takes the row for itself.
There is a test that asserts the strip stays one row.

## Verification

- `npm run lint` clean.
- 1179 tests across `src/tui/` + `src/config/` pass; the 5 failures are all present on pristine `main` (stale splash-banner copy, `tui-app` smoke, `persist-embedding-hybrid-recall`).
- New: 7 submit-routing tests, 4 Ctrl+T tests, 3 reducer tests, 3 config tests, 6 hint-strip tests.


sosidudku1:
The config work is done properly — schema field, default, a validating parser that throws on unknown values, and a persist helper that spreads both levels and re-validates, so it doesn't reintroduce the field-dropping class. I also checked the two races the feature invites and both are handled: the mode is read at call time with `state` in the dependency array, so a Ctrl+T flip mid-session takes effect on the next Enter; and if the turn ends between keypress and dispatch, `steerMessage` falls back to `sendMessage`.

**One inconsistency: bare `/steer` and `/queue` change the mode in memory only.**

`dispatchSteerSub` and `dispatchQueueSub` return a state action and a confirmation message, and `runSlashCommand` has no branch calling `onWhileBusyModePersistRequested` — while the sibling theme path directly above it does exactly that, and the Ctrl+T handler at `app-key-bindings.ts:127` persists too.

So the operator types `/queue`, sees "Enter now parks messages", and on the next launch `createInitialTuiState` seeds from `config.tui.whileBusySubmit` — still `steer`. The choice is silently gone.

The docstring is explicit that bare `/steer` is the persisting form, distinguishing it from `/steer <message>` which lands one message "without changing the persisted default". `app-key-bindings.test.ts:627` asserts the Ctrl+T persist; nothing covers the slash path.

plombeer31:
Fixed in e502747.

`dispatchSteerSub` / `dispatchQueueSub` now return `setWhileBusyMode` next to the `while_busy_mode_changed` action, and `runSlashCommand` forwards it to `onWhileBusyModePersistRequested` — the same callback the Ctrl+T binding calls, so bare `/steer`, bare `/queue` and Ctrl+T all land in the single `persistUserWhileBusySubmit` helper with its one try/catch in `tui-command.ts`. Shaped after the `setThemeName` branch directly above it; no second mechanism.

The message-carrying form is unchanged: `/steer <msg>` and `/queue <msg>` return `submitWhileBusy` and leave `setWhileBusyMode` unset, so they still land one message without moving the persisted default. `/queue clear` likewise.

Tests: `submit-handler.test.ts` asserts the persist callback fires with `"steer"` / `"queue"` for the bare forms (mirroring `app-key-bindings.test.ts:627`) and asserts it is *not* called for `/steer <msg>`, `/queue <msg>`, `/queue clear`; `slash-command-handler.test.ts` pins the result field on both sides. Reverting either half of the source hunk fails them (2 and 3 failures respectively).

`npm run lint` clean. `vitest run src/tui/`: 100/104 files, 1004/1009 tests pass — the 5 failures (splash-banner, chat-log, two tui-app smokes, `localModels.embeddings`) reproduce identically at stock 3bc6f69. AGENTS.md §Mid-turn steering updated to say all three routes share the persist path.

PR #160, opened 2026-08-19
feat(sidecar,http): steering API so hosts can redirect a running turn

plombeer31:
> **Stacked** on #157 — this branch contains that commit too. Merge
> it first; the commit that belongs to *this* PR is the last one.

## What this is

The runtime gained mid-turn steering in #157. This exposes it on the
two surfaces the desktop shell talks to.

Both handlers **deliberately bypass `turnController.enqueue`**.
Enqueueing would park the message behind the very turn it is meant to
redirect, which is the whole bug.

## Sidecar (NDJSON)

- `steer_message` request `{sessionId, text}` → `{steered}`. `false` when the session is not the active one or has no turn in flight — the host's cue to fall back to `send_message`.
- `steer_applied` event when a message reaches the model, carrying the step index, so hosts can render it inline in the running turn rather than as the start of a new one.
- `steer_undelivered` event, emitted from `send_message` for anything `RunTurnResult.undelivered` hands back. The sidecar has no queue of its own, so a late steer goes to the host rather than nowhere.

## HTTP (`atomic-agent serve`)

`POST /api/sessions/{id}/steer`, body `{text}`:

| status | meaning |
|---|---|
| `200 {steered:true}` | folded into the running turn |
| `409` | session is idle — the body names `/v1/chat/completions` as the right endpoint. Refusing beats accepting a message nothing will read. |
| `429` | per-session inbox full |
| `400` | missing or blank `text` |

## Verification

- `npm run lint` clean.
- 66 tests across `src/sidecar/` + `src/http/` pass. The one failure (`send-message-concurrency`) fails identically on pristine `main` — same assertion, same spurious `m3` from the reflection filter.
- New: 5 HTTP route tests (including the 429 cap boundary at exactly 16), 3 sidecar tests — the important one proves a steer resolves *while* a turn is blocked mid-inference, which a queued handler could not do.


sosidudku1:
The HTTP surface is correctly guarded — I checked this in the router rather than assuming it: the route omits `requiresAuth`, `compileRoute` defaults it to `true`, and only `/health` and `/v1/models` opt out explicitly. Input validation is genuinely thorough: body capped at 1 MiB, non-string/empty/whitespace text rejected, unknown and idle sessions both 409, and a full inbox 429s rather than dropping silently.

One blocker: **the HTTP path never consumes `undelivered`, so a steer it accepted with `200 {"steered":true}` can vanish.**

`grep -rn "undelivered" src/ | grep -v "\.test\."` returns six lines: three producers plus the type in `agent-loop.ts`, and two consumers in `sidecar/main.ts`. Nothing under `src/http/`. `openai-chat-completions.ts` reads `result.session` and never touches `result.undelivered`.

So when the turn ends before another step boundary, `flushSteering` sweeps the message into `undelivered` — and since the flush empties the inbox as it reads, once the HTTP caller ignores the value the text exists nowhere. The host has no way to detect the loss or resend.

The sidecar half of this same PR handles it:

```ts
for (const text of result.undelivered ?? []) {
  protocol.emitEvent("steer_undelivered", { sessionId, text });
}
```

with `sidecar-events.ts` stating the invariant outright — "Emitted rather than silently dropped so 'the message you sent always goes somewhere' holds on this surface too." The route's own docstring makes the same promise. It holds on one surface and not the other, in a PR that ships both.

plombeer31:
Fixed in 09d3545 — the HTTP path now consumes `undelivered` on every exit.

The trap I had to work around: the steer and the turn are different exchanges. `POST .../steer` was answered long before the turn closed, and the completion response goes to whoever owns the turn, who is generally not whoever steered. So bolting a field onto the completion response alone would have handed the message to the wrong party. The route that ran the turn (`openai-chat-completions.ts`) therefore **always** parks the hand-back in a per-session `UndeliveredSteerStore` — on the success path, the `failed` path, and the threw path (there nothing flushed the inbox, so it drains it rather than let a steer resurface inside an unrelated later turn) — and only mirrors it onto the response as a fast path.

How a host detects and resends:

1. `GET /api/sessions/{id}/steer` → `{sessionId, undelivered: [{seq, text, parkedAt}], discarded}`. This is the path for the host that sent the steer and holds no completion.
2. Resend is an ordinary `POST /v1/chat/completions` with that text.
3. `DELETE /api/sessions/{id}/steer?through={seq}` acks it.

Reads do not consume, deliberately: a retried or prefetched `GET` would otherwise lose the text, which is the failure mode this whole path exists to prevent. The ack takes a cursor from the listing, so a steer parked between the read and the ack has a higher `seq` and survives instead of being cleared unseen.

The host holding the turn does not have to poll: `undelivered_steers` on the non-stream `chat.completion` body (absent when the turn delivered everything, so vanilla completions are byte-identical), and `event: steer_undelivered` on extensions-opt-in streams — same name as the sidecar event; extensions-off streams stay strict OpenAI and use the `GET`. Both carry the parked `seq`, so the mirrored entry and the parked one are the same message, not two copies to resend twice.

Also: `DELETE /api/sessions/{id}` drops that session's parked steers with the row, and the per-session cap reports `discarded` rather than quietly returning a short list.

Tests: `route-sessions.test.ts` drives the real routes end to end — the steer goes in through `POST .../steer` during the final inference, the turn strands it, and the test asserts the inbox is empty afterwards, then retrieves, resends and acks it. Plus `openai-chat-completions.test.ts` (body field, SSE event, parked on the 500 path, field absent on a clean turn) and `undelivered-steers.test.ts` (non-destructive read, cursor ack, cap accounting). Checked they are not vacuous: reverting `openai-chat-completions.ts` to the previous commit fails 9 of them; making the `GET` consume fails 3; making the ack ignore its cursor fails 2.

`npm run lint` clean. `npx vitest run src/http/ src/sidecar/` → 84 passed, 1 failed: `send-message-concurrency`, which fails identically at stock HEAD with this diff stashed.

One gap I did not close, since it is outside this PR: `TaskRunner` also calls `runTurn` and drops `undelivered`, so a steer aimed at a task-driven turn can still vanish. Worth a follow-up — AGENTS.md now states the obligation for every surface that calls `runTurn` directly.

sosidudku1:
The main fix is solid: `parkUndelivered` consumes `undelivered` on all four exit paths, and the message goes into a real store readable at `GET /api/sessions/{id}/steer`, mirrored on the response and on the SSE event. That is delivery rather than logging, which was the point. The non-consuming GET plus a seq-cursor DELETE is also the right shape — it avoids dropping messages parked between the two calls, and both new routes inherit `requiresAuth` correctly.

Two things in the new store.

**1. `ack` destroys the loss counter (`undelivered-steers.ts`).**

```ts
if (box.entries.length === 0) this.bySession.delete(sessionId);
```

The cursor covers `entries`, but deleting the box also drops `discarded`. So a host that acks the highest seq it was given — before, or in the same pass as, inspecting `discarded` — permanently loses the "N messages were dropped" signal, and the next `GET` reports `discarded: 0`.

Confirmed by driving the store directly: overflow by 3, then ack the highest listed seq, and `discarded("s1")` returns 0 rather than 3. The host is now told nothing was lost when three messages genuinely were.

The comment above that line claims the opposite — "the discard notice included" — but nothing proves the host ever read it. Keeping the box alive while `discarded > 0`, or requiring a separate ack for the counter, would make the comment true.

**2. `park` reports only survivors when a single call overflows (low).**

`park` pushes the batch, splices the oldest out, and returns `parked.filter(entry => retained.has(entry))`. Since the response payload is built from that return, a `RunTurnResult.undelivered` batch larger than the remaining cap silently omits the oldest entries from the very response that is supposed to hand them back. The only trace is `box.discarded` — which, per the above, is itself erasable.

One cross-PR note rather than a defect: #157 removed the `isBusy` pre-check inside `steer()` as an unreliable fact, but this PR's `POST /steer` route still gates on `turnController.isBusy(id)` before calling it. After both land that check reintroduces the same stale read. It only rejects and never strands a message, so it's a reliability wrinkle rather than a correctness one — but worth reconciling when these two rebase.

plombeer31:
Both store defects confirmed by driving it directly before touching anything — `discarded` after acking the highest listed seq came back `0` instead of `3`, and a 19-message batch returned 16. Three commits (a994777, b328445, 74a133a).

**1. `ack` no longer erases the loss counter.** Both of your options, because one without the other is still wrong: keeping the box alive makes `discarded` permanently sticky (a polling host can't tell new losses from old), and an ack for the counter alone doesn't help if the box is already gone.

- `ack` touches entries only. `ackDiscarded(sessionId, count)` counts the loss down — a count, not a flag, so discards landing between the host's `GET` and its `DELETE` stay outstanding rather than being cleared unseen. Same cursor discipline as the entries, applied to a counter; the discarded messages have no `seq` the host was ever shown, which is exactly why the entry cursor can't stand in for having read the count.
- `DELETE /api/sessions/{id}/steer` now takes `?through={seq}` and/or `?discarded={n}`, at least one required (a bare `DELETE` is still 400). The response carries `discardsAcked` and the loss still outstanding, so a host that only ever calls `DELETE` sees the signal too.
- The comment now describes what the code does.
- Leak: a box outlives its entries only while a loss is unacknowledged, and is still reclaimed when both are empty, by `DELETE /api/sessions/{id}` (checked — that path is `clear()`, unaffected), and by `MAX_PARKED_SESSIONS` eviction. Pinned by a test that overflows, acks the entries, never acks the counter, and asserts the box is gone after eviction.

**2. `park` returns the batch whole.** Capacity is `max(MAX_PARKED_STEERS, batch.length)`, so eviction can only reach entries parked by *earlier* calls — never the batch just handed over. `park` returns it unfiltered and everything returned is retrievable until acked. That follows the cap's own stated purpose ("only bites when several turns strand messages and nobody ever acks"); the alternative, reporting the omission on the wire, would have meant a response listing entries the host can never fetch again. Not reachable from today's callers, as you noted, but the store owns its own constant and the failure mode if it drifts is a message that exists nowhere.

**3. `isBusy`.** Removed from the route. `steer()` is the authority; the route only translates. The inbox is read *after* a refusal and only to pick the status code (refusal + full inbox → 429, otherwise 409), so a stale read there can at worst mislabel a message that was definitively not queued, where the pre-check could reject one the runtime would have taken. 409-for-idle and 429-for-full are unchanged; the 409 text now says "no turn accepting steers" instead of "no turn in flight", which is the fact `steer()` actually reports.

This does not depend on #157 landing: it reads `steeringInbox.peek`, which both versions expose, and with today's `steer()` the pre-check is merely redundant. The one thing #157 changes is what `steer()` returns in the drain→`busy.delete` gap — refuse instead of accept — and the route reports that faithfully either way. If #157 ends up giving `push` a richer return than a boolean, the classifier here should read that instead of peeking, but nothing breaks if it doesn't.

`npm run lint` clean. `vitest run src/http/ src/sidecar/ src/cli/ src/runtime/ src/agent/`: 311/312, the one failure being `send-message-concurrency`, which fails identically at stock HEAD here. Each new test proved non-vacuous by reverting its hunk: restoring the `bySession.delete` drops 3 store tests plus the route-level one (`discarded: 1` → `0`); restoring the survivors filter drops 2; restoring the `isBusy` gate turns the new POST test into `expected 409 to be 200`.

PR #165, opened 2026-08-19
feat(tui): mouse support — click the nav bar, panels, selectors and the prompt

plombeer31:
## What

Makes the TUI clickable: the navigation bar and sub-tabs, the sidebar, every list panel, the pickers and modals, tool cards, and the prompt input (click to place the caret). The wheel scrolls the chat and walks the focused panel.

## Why this reverses a deliberate decision

`src/tui/tui-command.ts` carried an explicit comment choosing **not** to enable SGR mouse tracking, because capture takes away the terminal's own drag-to-select — and Apple Terminal has no Shift-bypass. That reasoning is sound, but it is an argument for a **setting**, not for having no mouse layer at all. So:

- mouse reporting is **on by default**, and
- there are three ways to turn it off: `/mouse off` at runtime, `atomic-agent tui --no-mouse` for one run, and `"tui": { "mouse": false }` in `config.json` (config v38, transparently upgraded).

With mouse off, behaviour is byte-for-byte what it was before: alternate-scroll (`\x1b[?1007h`) turns the wheel into cursor keys and the old arrow-key chat scroll handles it.

Only **1000 + 1006** are requested — button events plus SGR coordinates. Motion tracking (1002/1003) is deliberately not enabled: nothing in the UI hovers or drags, and motion reports are a constant wakeup stream for a UI that would ignore them.

## How it works — `src/tui/mouse/`

| File | Role |
|---|---|
| `mouse-tracking.ts` | Enable/disable the reporting modes. Same shape as `alt-screen.ts`, including the `process.on("exit")` restore so a crash never leaves the host terminal printing click escapes into the user's shell. |
| `parse-mouse-events.ts` | Pure decoder for SGR and legacy X10 reports. Reassembles a report split across two reads. A lone trailing `ESC` is passed straight through — buffering it would delay the Escape key until the next keystroke. |
| `mouse-stdin.ts` | Ink has no mouse layer and would type the raw reports into the chat buffer, so Ink is handed a `PassThrough` carrying only keyboard bytes, with `isTTY` / `setRawMode` / `ref` / `unref` proxied to the real stdin. |
| `mouse-registry.ts` | Hit testing. Ink exposes no absolute positions (`measureElement` returns a size only), but every node keeps its Yoga node — `absoluteRect` sums `getComputedLeft/Top` up the parent chain, the same walk `render-node-to-output.ts` performs when painting, so the rectangle is exactly where the node was drawn. Ancestors with `overflow: hidden` clip it. |
| `mouse-context.tsx` | React glue: a provider carrying `dispatch` / `callbacks` / a live `getState`, plus `useMouseTarget` and a layout-neutral `<MouseTarget>`. Outside the provider (component tests, `--no-mouse`) every clickable component renders exactly as before. |
| `mouse-list-row.tsx` | The shared list row: click selects, a second click on the selected row activates. |
| `synthetic-key.ts` | Activation and the wheel are fed to each panel's existing `*-key-bindings.ts` handler as a synthetic Enter / arrow key. |

Two behavioural rules worth calling out:

- **First click selects, second click activates.** No double-click timing window (unreliable over SSH, invisible to the user) and no open-on-first-click, which would make a mis-click destructive in lists where Enter starts a download or switches sessions. It mirrors what the keyboard already does: arrow to the row, then Enter.
- **A modal owns the mouse exactly when it owns the keyboard.** `TuiApp` raises the registry's layer floor using `isPanelModalOpen` — the same predicate extracted out of `handleAppKey` — so a click can never reach the list rendered behind an open confirm.

## What became clickable

- Run / Observe / Manage pills and the Observe / Manage sub-tab strip
- Sidebar: session rows and task rows (click focuses the rail and moves the cursor; click again opens)
- Lists: skills, skills hub, tasks, memory, MCP, LLM panel (all four panes), providers, local models
- Pickers: session, theme (with live preview on select), slash palette
- Approval modal `[y]` / `[s]` / `[a]` / `[n]`, plus the `y` / `n` hotkey chips
- Hotkey chips with an unambiguous meaning (`tab`, `shift+tab`, `esc`, `/`) — chips that name a gesture rather than a command stay plain text
- Tool cards: the per-card expand toggle existed in the reducer but had **no key binding at all** — only `/expand` and `/collapse`, which act on every card. The mouse is the first way to open one card.
- The prompt: click places the caret. `rowColToCursor` does not clamp, so a click past the end of a short line would have run the offset into the next line; it is clamped at the call site, making a click in the empty space right of a line mean "end of this line".

## Supporting changes

- Absolute `*_cursor_set` actions for the cursors that were delta-only: sidebar sessions/tasks, session picker, theme picker, providers, local models, skills hub.
- `isPanelModalOpen` and `decideApproval` extracted from `app-key-bindings.ts` so the mouse and the keyboard share one predicate and one approval path instead of two copies that can drift.
- The nav pill, sub-tab and hotkey strips became per-item boxes (they were single `<Text>` runs) so each item is measurable; they use `flexWrap="wrap"` so a narrow terminal wraps whole chips instead of clipping characters.
- `tui.mouse` config key (v38), `--mouse` / `--no-mouse` flags, `/mouse [on|off]` command.
- README + AGENTS.md document the layer and the selection trade-off.

## Testing

- `npm run lint` clean.
- New unit tests: escape sequences (`mouse-tracking.test.ts`), the decoder including split chunks / wheel / modifiers / X10 / lone-ESC (`parse-mouse-events.test.ts`), the stdin split (`mouse-stdin.test.ts`), and hit testing including nesting, clipping and the modal layer gate (`mouse-registry.test.ts`).
- `mouse-app.test.tsx` drives the real Ink tree: it finds a label in the rendered frame and emits a click at those coordinates, then asserts the section switched, the sub-tab switched, the caret landed mid-word, a click past EOL clamped, the wheel walked a panel cursor, and a row click moved the cursor. Because Ink commits frames on a ~30fps throttle, these poll the frame instead of sleeping a fixed interval.
- Full suite on this machine: **8 failed / 4113 passed**, against **9 failed / 4063 passed** on `main` at `667dae1`. Every failure on the branch is also a failure on `main` (stale splash/hint expectations, the `$HOME`-dependent `fs-glob-real` case, and the known flaky `send-message-concurrency` / `llm-health-poller` pair) — no new ones, +50 tests.
- Driven manually in a real PTY against the built binary (`node dist/cli/index.js tui`) with synthetic SGR reports, rendered through a terminal emulator: mouse enable sequences on startup ✅, click on the Observe pill switches section ✅, click on the Logs sub-tab switches tab ✅, click mid-word in the prompt puts the caret there (typing produced `helXlo world`) ✅, `/mouse off` writes `\x1b[?1006l\x1b[?1000l` and confirms on screen ✅, `/mouse on` re-enables ✅, quitting restores reporting **and** leaves the alt screen ✅.

## Overlap with the open run-mode stack (#161–#163)

Branched off `main @ 667dae1`, so it is conflict-free today, but two touch points are worth knowing:

- **Config version.** This PR takes `USER_CONFIG_VERSION` 37 → 38 for `tui.mouse`; #161 takes the same step for `llm.runMode`. Whichever lands second should rebase onto 39 — the parser change is a one-line `SUPPORTED_INPUT_VERSIONS` entry plus the constant.
- **TUI chrome.** #163 adds a Run submenu row to `TuiApp`; this PR splits the status-bar pills and the `SubTabBar` strip into per-item boxes. Both edit the same region of `status-bar.tsx` / `debug-pane.tsx`. Rebasing this one after #163 is the easier direction — the new submenu row just needs the same `<MouseTarget>` wrapper the other strips got, and it becomes clickable for free.


sosidudku1:
The terminal-mode hygiene here is genuinely good and worth crediting: cleanup runs on normal quit, SIGINT/SIGTERM/SIGHUP and via a `process.once("exit")` net, `disable()` is idempotent, SGR 1006 is negotiated so wide terminals work, hit regions unregister on unmount, and the text-selection tradeoff is acknowledged and mitigated by making it a toggle rather than ignored.

One blocker: **`/mouse on` is half-wired.**

`tui-command.ts:208` computes `mouseEnabled` once at startup, and line 444 gates the prop on that value:

```ts
...(mouseEnabled ? { mouse: mouseSource } : {}),
```

`setMouseEnabled` reassigns the local `mouseTracking` controller and persists the setting, but never re-renders or otherwise informs the tree. So starting with `tui.mouse: false` and then typing `/mouse on` writes the escape sequences and reports "mouse support on — click panels, rows and the prompt" while the mounted `TuiApp` still has `mouse === undefined` and its effect returns early at `if (!mouse) return;`.

Net result: clicks do nothing, drag-to-select is now broken, and the UI claimed it worked. Only a restart fixes it.

Since `mouseSource` is already created unconditionally and is inert when tracking is off, passing it unconditionally should be enough.

plombeer31:
Fixed in a34db18.

**The fix.** `mouse: mouseSource` is now passed unconditionally. Your one-liner was the right diagnosis, but I added a second half: the stdin forwarder is now `(event) => { if (mouseTracking) mouseSource.emit(event); }` instead of a bare `mouseSource.emit`. It reads the `let` binding per report rather than capturing it, so the tracking controller — not a mount-time boolean — is the single live gate. Without it `/mouse off` only stopped clicks because the terminal went quiet; anything that still arrived (a multiplexer that swallowed the disable, a paste carrying an SGR report) would have been acted on.

**Live pty check**, `tui.mouse: false` in the config, real TUI, `pyte` screen, SGR reports injected only while the app has actually asked for reporting:

| step | before | after |
|---|---|---|
| boot | no `1000h` written | no `1000h` written |
| click Observe pill (mouse off) | terminal sends nothing, no switch | same |
| `/mouse on` | writes `1000h` `1006h`, banner shown | same |
| **click Observe pill** | **nothing happens** | **`▸ Observe`** |
| click Run pill | — | `▸ Run` |
| `/mouse off` | writes `1006l` `1000l`, banner shown | same |
| click Observe pill | no switch | no switch |
| forced SGR bytes with mouse off | ignored (no subscriber) | ignored (forwarder gate) |
| `/mouse on` again → click | **nothing happens** | **`▸ Observe`** |

Mouse bytes never leaked into the chat buffer in any state.

**Tests.** New `src/tui/tui-command.mouse.test.ts` boots `tuiCommand` with a fake tty and a mocked `render`, captures the props handed to `TuiApp`, and drives `onMouseSupportRequested` exactly as the slash command does — asserting that bytes pushed through stdin land on the source the tree subscribed to at mount. Not vacuous: reverting the unconditional prop fails 3 of the 5, reverting the forwarder gate fails 2. Together with `mouse-app.test.tsx` (source event → real Ink tree → UI moves) the path from a byte on stdin to a section change is covered end to end.

One incidental: `tui-command.ts` imports `node:sea`, which Node only publishes under the `node:` prefix while Vite's builtin check strips it — so importing this module in a test is unresolvable without `vi.mock("node:sea", …)`. Mocked locally with a note rather than touching `vitest.config.ts`.

`npm run lint` clean. `src/tui/mouse/` + the new file: 46 passed, 0 failed. Full suite (minus the `src/tools/os/git/*` suites, which time out in my sandbox): 4099 passed, 8 failed across 6 files — the same 8 that fail at 78a8bc0 unmodified (splash-banner/chat-log and tui-app fixtures, `localModels.embeddings.url`, a dev-machine `fs-glob` path, `send-message-concurrency`).

Also added an AGENTS.md line under §Mouse support recording why the prop is unconditional, since the failure mode is invisible until you toggle at runtime.

sosidudku1:
Rebased this branch onto current `main` and force-pushed the result, so this PR is live again rather than being superseded by a new one. Your original commits are preserved on `backup/pr165-original` (`a34db18`) if you want to compare or recover anything.

Nothing about the mouse engine changed. `src/tui/mouse/` — the SGR/X10 decoder with split-read reassembly, the Yoga-based hit testing, the stdin split that keeps Ink from typing raw reports into the chat buffer, the layer gate — merged onto current `main` without a single conflict, which is a good sign for how it was factored. All 15 files came across as-is.

What did need rewriting is the chrome, because `main` moved underneath it:

**The pill strip is gone.** #170 replaced the Run / Observe / Manage pills and the sub-tab strip with a breadcrumb plus one menu registry. Rather than reinstate the pills — which would give navigation two competing controls — the breadcrumb itself now takes the click and opens the menu, exactly as `ctrl+p` does. `/mouse` is registered as a menu node (`setup.mouse`), so the slash command and the menu entry come from one definition instead of two.

**Three places where `main`'s newer decision won over the branch:**

- The hint strip keeps its one-row shedding contract. This branch switched it to `flexWrap`, which produces exactly the two-line smear the `shed` ranks were added to prevent. Chips became individually measurable boxes instead, so hit testing works and the ranks stay intact.
- `ctrl+p` keeps the slot the branch gave back to a `/ commands` chip. `main` removed that chip deliberately when `/` stopped opening the palette over a non-empty draft.
- `USER_CONFIG_VERSION` goes to **40**, not 38 — `main` reached 39 (the `web.fetch` block) after this branch was cut. 38 stays in the supported list.

**One bug found while porting.** `SidebarRow` passed a window-relative row index to `sidebar_tasks_cursor_set`, which takes an absolute one. In a scrolled task rail, clicking a row selected a different task. Both lists now pass `window.start + idx`.

**Tests.** 4766 passing, zero new failures against `main`'s 5 pre-existing ones (3 stale TUI smoke expectations, 1 flaky `send_message` concurrency case, 1 `$HOME`-dependent glob test). The two tests that drove the deleted pill and sub-tab strips are replaced by one that drives the breadcrumb; I checked it fails if the click handler is removed, so it is testing the path and not just the frame.

Two notes on the PR description, since it no longer matches the branch: the config version is 40, and the "Overlap with the open run-mode stack" section is now moot for #163 (the fusion stack is still open and unmerged, so the config-version note there still applies to whichever of us lands second).

Happy to adjust any of the three chrome calls above if you disagree — particularly the breadcrumb-opens-menu choice, since that is the one that changes what your PR does rather than just where it sits.

PR #166, opened 2026-08-19
feat(tui): verify a cloud API key is active and funded before saving it

plombeer31:
> Stacked on #152 — based on `fix/wizard-empty-api-key`, so review that one first. GitHub will retarget this to `main` when it merges.

## Problem

A cloud key that is present is not a key that works. Any non-empty string is written to `.env` and the provider is saved **and made the active text provider** without a single call to the service. Nothing in the repo ever verifies a cloud key: there is no authenticated probe for `openrouter` / `aimlapi` / `gemini`, and no balance check anywhere. A revoked key, a key pasted from the wrong service, or a key on an account with no credit all configure cleanly and fail on the first real message, in the middle of a turn, as an unexplained provider error.

Listing `/v1/models` does not settle it either — plenty of endpoints list models for an unauthenticated caller, and no listing is ever billed. The only answer that proves both authentication *and* funds is a completion.

## Changes

**`src/llm/provider/verify/`** — UI-free, so a later "test key" action on the Providers tab can reuse it verbatim:

- `verifyProviderKey(target)` posts one `max_tokens: 1` completion via `openAiFetch` — deliberately *not* `openAiPostJson`, whose 3-attempt retry budget and `retry-after` waits are wrong for an interactive check. 8s timeout, at most 3 requests total (model → other token field → next model), cancellable, and the key is redacted from anything it reports.
- `classifyVerifyResponse` maps the answer onto `ok | invalid_key | no_balance | model_unavailable | rate_limited | unreachable | timeout | provider_error | cancelled`. Status codes alone do not settle it: prepaid services answer 401/403 once credit runs out, OpenAI sends `429 insufficient_quota`, and Gemini answers **400** for a bad key — so the body is read for billing/key wording first. A bare 429 stays soft: throttling proves the key authenticated.
- `pickProbeModels` spends the token on the cheapest **paid** OpenRouter model (~$0.000001 a check, from the catalog's own pricing). Free models and `openrouter/auto` are excluded on purpose — they answer 200 on a key with zero balance, which would make the balance check a formality. Where a catalog carries no prices (AI/ML API, Gemini, arbitrary compatible endpoints) the model the operator just chose is the probe.

**Wizard integration** — `verifyWizardBeforeSave` is the single seam, shared by `ProvidersOrchestrator.completeWizard` and first-run `CloudProviderOnboarding`, so the two save paths cannot drift:

- `invalid_key` and `no_balance` **block**: nothing is written to `.env`, nothing is persisted, nothing is activated, and the wizard stays open with a sentence saying which it was.
- unreachable / timeout / rate-limited / no-usable-model **save with a visible warning**. An offline laptop or a machine behind a proxy has to stay configurable; the key is simply reported as unverified (onboarding prints its warning after the Ink tree is torn down).
- Esc cancels a check in flight — it is the one key that survives the submitting lockout — and a verdict arriving after the cancel is dropped rather than overwriting the reopened wizard.
- Nothing is probed for services that legitimately have no key, or for a base URL on this machine (`is-local-provider-url.ts`, now shared with the startup gate instead of duplicated).

Out of scope on purpose: switching to an already-configured provider (no network round-trip per keypress), bulk config import, and hand-edited `config.json`.

## Tests

`classify-verify-response.test.ts` (one case per rule, including `403 insufficient credits` → no_balance and Gemini's 400), `verify-provider-key.test.ts` (401 answers in **one** request; 402; 404 → second candidate; `max_tokens` → `max_completion_tokens` resend; timeout; cancellation; the key never reaching `detail`), `pick-probe-models.test.ts` (a free row can never be chosen), `verify-wizard-before-save.test.ts` (blocking vs warning; local providers never touch the network; Gemini's `/v1beta/openai` path), and `completeWizard` cases in `providers-orchestrator.test.ts` (a 402 emits `providers_wizard_failed`, never `succeeded`, and never reloads the registry — proof nothing was saved).

`npm run lint` clean.


sosidudku1:
The security design is right where it matters, and I checked it specifically: the key goes only to the provider's own catalog-derived endpoint, the request is bounded by an 8s timeout, error detail is redacted and length-capped, and nothing reaches the analytics client. Only `invalid_key` and `no_balance` block a save — timeout, unreachable, rate-limited and model-unavailable all save with a warning, so an offline operator can still configure a key. That last decision is the right call.

One blocker, in the onboarding path only.

**`cloud-provider-onboarding.tsx:43` — a cancelled verification can still save.** After the await, the only guard is `if (!alive.current) return;`, which is a mount check — and Esc doesn't unmount the component. `verifyProviderKey` samples `signal.aborted` at the top of each loop iteration and in the fetch catch, so if the abort lands after the response arrives but before `classifyVerifyResponse` returns, the function returns a normal verdict rather than `"cancelled"`. Execution then falls through to `saveProviderWizardToConfig(nextWizard)` and `onFinished("saved_cloud")`.

The operator sees "Key check cancelled — press Enter to try again" while the provider is written to config and onboarding exits as saved.

The orchestrator already handles this exact case, with a comment naming the hazard:

```ts
const gate = await verifyWizardBeforeSave(wizard, { signal: abort.signal });
// A cancel already put the wizard back in an editable state; a
// late verdict from the abandoned check must not overwrite it.
if (abort.signal.aborted) return;
```

So the fix is one line in the onboarding twin, mirroring the sibling.

**Related, same file:** `submit` guards re-entry on the `submitting` state captured in its closure, but the cancel handler resets that state — so Enter after Esc starts a second verification while the first is still live. Combined with the above, the abandoned call can win the race to save. A generation counter or an in-flight ref that survives the state reset would close both.

All 169 tests in `src/llm/provider/verify` and `src/tui/providers` pass; there's no coverage of cancel-then-resolve, which is why this slipped through.

plombeer31:
Both fixed, plus the base merge.

**Base merged first.** `fix/wizard-empty-api-key` @ `3da0194` is in as its own commit (`4fdcb84`). Git auto-merged it, and the result keeps both sides: `apiKeyForWizard` still returns early on the typed buffer and still falls back to `storedEntryForWizard` → `resolveLlmProviderApiKey`, so a `configure` run whose key lives in `config.json` resolves it — and `verifyTargetForWizard` reads through the same function, so that stored key is what gets checked. `isWizardFirstScreen`, the Esc handling and `emptyKeyMeaningForWizard` came over unchanged. Lint clean and 289 providers/verify tests green on the merge commit alone, before any new code.

**Cancel.** Guarded, but not with the one-liner as written — `alive.current` is now `checkStillWanted(abort)` = `alive.current && !abort.signal.aborted`, applied at both exits of `submit`.

**Re-entry.** This is where the one-liner is not sufficient, and it is worth being precise about why. The `submitting` closure guard is gone; re-entry is guarded on the in-flight `AbortController` ref, which is written before the first await and cleared only by the run that owns it or by a cancel.

Enter-after-Esc alone would have been covered by the abort check, since the cancel aborts. What is not covered is two key events drained from stdin in one turn — `handleReadable` loops `while ((chunk = stdin.read()) !== null)` and emits each event synchronously, and both read the same stale `submitting: false`. **Neither run is cancelled in that interleaving**, so no post-await signal check can separate them. Measured against the unfixed component: 2 probe requests, 2 `saveProviderWizardToConfig` calls, 2 `onFinished("saved_cloud")` calls. The ref is what closes it; the abort check then settles the rest, because with the ref there is at most one non-cancelled run alive.

The `catch` exit got the same guard. It is not reachable from a provider answer today — `verifyProviderKey` turns every transport failure into a verdict rather than a rejection — but an abandoned run that did throw would paint over the screen the operator was handed back and free `submitting` under a live check, which is the same bug at the other exit.

**Coverage** — `src/tui/components/cloud-provider-onboarding.test.tsx`, 5 cases, driving real key bindings and the real verify path. The fetch stub answers immediately but parks `res.text()`, which puts the abort exactly where you described it: after the response, before `classifyVerifyResponse`.

- Esc mid-check, verdict arrives after → nothing written, `onFinished` not called, cancel notice still on screen
- Esc then Enter, abandoned check answers last → it cannot save; the retry does, once
- two Enters in one stdin drain → one probe, one save, one `onFinished`
- abandoned check *rejects* after the retry took over → its message never lands, `submitting` stays owned by the retry
- Ctrl+C mid-check, verdict arrives after → nothing written

Not vacuous: reverting each of the three hunks separately fails only its own cases — re-entry ref → 1 failure (the double-Enter one); try-exit guard → 3; catch-exit guard → 1.

`npm run lint` clean. `npx vitest run src/llm/provider/verify/ src/tui/providers/ src/tui/components/cloud-provider-onboarding.test.tsx` → 186/186 in 15 files. Full `src/tui/` → 1013 passed, 5 failed, all 5 pre-existing at stock HEAD (splash-banner + chat-log + tui-app fixtures, `localModels.embeddings.url`); stock actually fails 6 there, the extra `tui-app` case being the known flake.

`AGENTS.md` gained the invariant and the new pinning test in the credential-check section.

PR #167, opened 2026-08-19
feat(llm): more cloud providers and a refreshed model catalog

plombeer31:
## Problem

Three separate things kept cloud models out of reach:

1. **No preset for most vendors.** `PROVIDER_PRESETS` covered 11 endpoints. Anthropic — the one vendor with no route into the agent at all — was not among them.
2. **A vendor filter in the live catalog.** `scoreChat()` in `fetch-openrouter-chat-catalog.ts` returned `-1` for every `anthropic/*` id and everything matching `/gemini/i`. That is ~40 currently served models (the whole Claude 5 and Gemini 3.x lines) deleted from the picker, with no operator override. Three tests pinned it; no comment anywhere said why.
3. **Stale bundled catalogs.** OpenRouter carried 18 chat rows, AI/ML API 13, and neither had a single Claude or Gemini row.

## What changed

**Eight new vendor presets** — Anthropic, Hyperbolic, Moonshot AI (Kimi), Novita AI, Perplexity, Qwen (DashScope), SambaNova. Each resolves to the existing `openai-compatible` kind with `baseUrl` prefilled, so there is **no new provider kind and no config-schema change**; model lists still come from each server's own `/v1/models`.

Every base URL was probed live, at the URL the code actually builds (`normalizeOpenAiBaseUrl(base) + "/v1/models"`):

| Preset | base URL | probe | control (`/v1/__nope_zzz`) |
|---|---|---|---|
| Anthropic (Claude) | `https://api.anthropic.com` | 401 | 404 |
| Hyperbolic | `https://api.hyperbolic.xyz` | 401 | 404 |
| Moonshot AI (Kimi) | `https://api.moonshot.ai` | 401 | 404 |
| Novita AI | `https://api.novita.ai/openai` | 200 `data[]` | — |
| Perplexity | `https://api.perplexity.ai` | 401 | 404 |
| Qwen (DashScope) | `https://dashscope-intl.aliyuncs.com/compatible-mode` | 401 | 404 |
| SambaNova | `https://api.sambanova.ai` | 200 `data[]` | — |

The control column is the part worth keeping: a 401 only proves the path exists if a bogus sibling path on the same host answers 404. **Dropped for failing that bar rather than shipped half-working:** z.ai and Cohere's compatibility root (both answer the same auth error for every path), and DeepInfra (its list lives at `/v1/openai/models`, which the "root without `/v1`" convention cannot express).

**The Anthropic/Gemini exclusion is lifted.** `scoreChat` now ranks vendors instead of gating them — Claude 5 and Gemini 3.x get positive scores, so the models this agent is tuned for still sort to the top, but nothing is hidden. A negative score is reserved for rows that genuinely cannot be used (non-chat surfaces, models that advertise no tool support). **This reverses an existing decision** — if it was deliberate policy rather than an accident, say so and I will drop this commit's third hunk; everything else stands on its own.

**Both bundled catalogs regenerated from the vendors' public endpoints** (2026-08-19). OpenRouter: 18 → 55 chat rows, with `context_length`, `input_modalities` → vision, `supported_parameters` → tools, `input_cache_read` → prompt cache, and price × 1e6 → USD per 1M. AI/ML API: 13 → 32, including the vendor-prefixed `anthropic/claude-*` and `google/gemini-*` ids it serves on the `openai/chat-completions` surface (the unprefixed legacy spellings stay retired — those really are gone).

The OpenRouter chat rows move into two sibling modules (frontier / open-weight) and the two copies of the row builders collapse into `model-catalog-entry.ts`, keeping every file inside the 300-line rule.

## Tests

- The three tests that pinned the exclusions now pin the opposite, each with a comment explaining the reversal.
- New: every chat row has a positive context window and a price; picker order has no duplicates and matches the map exactly (the split into sibling files makes a copy/paste duplicate otherwise silent); the new presets carry their own env var and the keyless ones stay flagged.
- `src/tui/llm-panel/llm-panel-selectors.test.ts` expected `qwen/qwen3.7-max` at `$1.25/$3.75`; OpenRouter charges `$1.475/$4.425` today, so the expectation moved to `$1.48/$4.42`.

`npm run lint` clean. `npx vitest run src/llm src/tui` green. Full `npm test` has 18 failures that also fail on a clean `main` in this environment (git-tool tests, a `$HOME` glob test, ink render smoke tests) — none in the touched area.


sosidudku1:
The catalog work checks out — catalog map and picker order are exactly in sync, no duplicates or orphans, and a config pinning a retired model id degrades through `resolveModel` to `DEFAULT_CHAT` rather than crashing. The catalog only feeds `resolveModelPricing`, never the prompt, so the stable prefix is untouched.

One blocker: **the Anthropic preset can't authenticate.**

The preset resolves to the `openai-compatible` kind, which sends `Authorization: Bearer` in both request paths — `fetch-openai-compat-models.ts:58` and `openai-http.ts:111`. `api.anthropic.com` requires `x-api-key` plus a mandatory `anthropic-version` header, and ignores `Authorization` entirely.

There's no way to correct it from config either: `ProviderPreset` carries only `id`/`label`/`baseUrl`/`envVar`/`listsModelsWithoutKey`/`local`/`note`, and `deps.extraHeaders` is populated solely from `options.headers`, which the wizard never sets.

So an operator picks the first row in the list, pastes a valid `sk-ant-…` key, and model discovery 401s — as does every subsequent chat turn. Because `presetId` is set, `nextPhaseAfterApiKey` skips the `base_url` step, so `baseUrl` is locked and there's no manual escape.

Worth flagging how this passed the file's own admission bar, which is written just above: a preset qualifies if `<baseUrl>/v1/models` answers "200 with a `data` array, or 401/403 while a bogus sibling path on the same host answers 404". Anthropic returns 401 here *because the auth header is wrong* — the exact reason that should disqualify it. The 401 proved the path exists, not that the preset can ever authenticate. `provider-presets.test.ts` only asserts the literal baseUrl and envVar strings, so nothing catches it.

Either drop the preset from this PR, or add a header-override field to `ProviderPreset` and thread it through `buildOpenAiHeaders`.

plombeer31:
Fixed rather than dropped — the preset now authenticates. Two commits on `valeryb/more-cloud-models`.

**Override design.** `ProviderPreset` gains two optional fields: `apiKeyHeader` (the header the key rides in; absent keeps the Bearer default) and `headers` (static headers the service mandates). Anthropic sets `x-api-key` and a pinned `anthropic-version: 2023-06-01`. The next non-Bearer vendor is a data change here, not a code change.

Both request paths now go through one builder, `openai/openai-auth-headers.ts`. That is the point of the file: they hard-coded `Authorization: Bearer` independently, so wiring one and forgetting the other was the available bug. `buildOpenAiHeaders` and `fetchOpenAiCompatModels` both call it and cannot drift. Keyless servers still send no auth header at all; static headers go out even without a key, since a version header is part of the request contract and not part of the credential. Header names normalize to lower case, which moved two existing `Authorization` assertions to `authorization`.

**Round trip.** You were right that this is where it would rot. The override lives on the saved entry, not on the build-time preset table: `buildProviderEntryFromWizard` copies both fields onto the `UserLlmProviderEntry` next to `apiKeyEnvVar`, `parseLlmProviderEntry` reads `apiKeyHeader` back out of `config.json` (`headers` already parsed), and `registerBuiltInProviderKinds` hands both to `OpenAiProvider`. So it survives restart, and a hand-written entry can express the same thing without a preset. The key itself stays out of config — it still resolves from `ANTHROPIC_API_KEY`, only the header *name* is stored. `deps.extraHeaders` was indeed the seam: the wizard now populates it.

**Corrected admission rule** (`provider-presets.ts` header comment + AGENTS.md). Probe `<baseUrl>/v1/models` **with the exact headers the preset will send**, and accept only: 200 with a `data` array, or a 401/403 that rejects the *credential* — body complaining about the key. A 401 whose body names a header the preset does not send (`x-api-key header is required`, `Invalid bearer token` for what is an API key) is now a **failing** probe. That is exactly the signal the old wording read as a pass. The bogus-sibling-404 clause is unchanged and still required in both branches.

Anthropic passes the new bar under the new headers — verified live, unauthenticated, no real key involved:

```
x-api-key: <bogus> + anthropic-version  → 401 {"message":"API key is invalid."}   (rejects the credential)
/v1/no-such-endpoint                    → 404                                     (sibling path)
```

and fails it under the old ones, which is the whole point:

```
no headers                              → 401 {"message":"x-api-key header is required"}
Authorization: Bearer sk-ant-…          → 401 {"message":"Invalid bearer token"}
```

The Bearer line is the load-bearing one: `api.anthropic.com` treats `Authorization: Bearer` as the OAuth path, so an `sk-ant-` key sent that way never reaches the key check on any path — the preset could not have worked with any key.

**What the header test pins.** `provider-presets.test.ts` asserted literal `baseUrl`/`envVar` strings, which is why this shipped. It now pins the bytes that leave the process, on both paths, after the entry has been through `JSON.stringify` → `parseLlmProviderEntry`:

- discovery — `fetchOpenAiCompatModels(baseUrl, key, preset)` sends `x-api-key` + `anthropic-version`, and `authorization` is absent;
- chat — the entry goes through the real `openai-compatible` registry factory into `OpenAiProvider.complete()`, and the captured request has `x-api-key` + `anthropic-version` and no `authorization`;
- the round trip itself — `apiKeyHeader`/`headers` survive config, and `apiKey` stays `undefined` so the secret is not written to config;
- every other preset is asserted to carry neither field, so a stray override can't silently break a Bearer-only vendor.

Plus unit coverage on `buildOpenAiHeaders` (Bearer default, override drops `authorization` entirely rather than adding alongside, keyless emits neither) and on the discovery fetch.

**Not vacuous** — each source hunk reverted in turn, against the three suites (52 tests):

| reverted hunk | failures |
|---|---|
| `buildOpenAiAuthHeaders` ignores `apiKeyHeader` | 5 |
| Anthropic preset loses both fields | 4 |
| wizard stops copying onto the entry | 2 |
| registry stops forwarding `apiKeyHeader` | 1 |
| discovery fetch back to hard-coded Bearer | 4 |

Restored: 52/52 pass.

**Verification.** `npm run lint` clean (`tsc --noEmit`). `npx vitest run` over `src/llm/provider/` + `src/tui/providers/` + `src/config/`: **0 failed**. Full suite: **4079 passed, 10 failed / 425 files**. Eight of those are exactly the known baseline — `send-message-concurrency`, `persist-embedding-hybrid-recall`, `tui-app` ×3, `chat-log`, `splash-banner`, dev-machine `fs-glob-real` — and I confirmed them at stock `2afd589`. The other two are `llm-health-poller`, a timer test that passes in isolation on both my branch and `2afd589` and flakes only under full-suite load; my diff touches nothing it reads. Nothing attributable to this change. `src/tools/os/git/*` happened to pass this run.

No conflict expected with #168.

PR #168, opened 2026-08-19
feat(models): ranked multi-term model search, in the TUI and the CLI

plombeer31:
## Problem

Model search was a single case-insensitive `includes` over the model id — `filterModelIds` in `providers-panel-state.ts`, shared by the `/model` modal picker and the Cloud pane's inline list. That is enough for the 18 rows the bundled catalog used to hold and useless against the 300–400 the live OpenRouter list returns:

- `claude vision` matches nothing — it is not a substring of any id.
- There is no way to ask for "cheap", "1M context", or "has prompt caching".
- Cloud models are not searchable from outside the TUI at all, which is exactly when you need them: picking a `defaultChatModel` for a config file, or checking what a provider charges.

## What changed

**One scorer, `src/llm/provider/model-search.ts`:**

- Query splits on whitespace; **every** term must match, so each word narrows.
- A term matches the id, the vendor prefix, or a tag derived from the catalog entry: `vision`/`text`, `tools`, `cache`, context shorthand (`1m`, `200k`), `free`/`cheap`/`routed`. Price tags mirror the *rendered* price, so `openrouter/auto` is `routed` and never `free` — you can search for what you can see.
- A term that matches no substring still matches as an ordered **subsequence**, ranked strictly last, so typos land somewhere.
- Results are ranked: exact id > id prefix > vendor > word start > substring > subsequence. **Equal ranks keep input order** — this runs on every keystroke and the curated catalog order has to survive, or rows jitter under the cursor.

**Both TUI surfaces keep working unchanged.** `filterModelIds` keeps its name and signature and delegates, gaining an optional catalog lookup so the Cloud pane can match on capabilities for curated kinds. No new keybindings, no new state: `f`, `cloudModelFilter` and the existing reducer actions carry it, and the pane's `no match` / `n/m of N` counter comes along for free.

**New CLI surface:**

```
atomic-agent models search <query> [--provider <id>] [--limit <n>] [--json] [--refresh]
```

```
$ atomic-agent models search qwen vision
or   qwen/qwen3.6-35b-a3b   262k · $0.15/$1 · vision · tools
or   qwen/qwen3.6-flash     1M · $0.19/$1.13 · vision · tools
```

It searches every configured provider's catalog, plus its `userModels`, plus — with `--refresh` — the live lists (`kimi` goes from 1 hit to 8). Rendering reuses the picker's formatters, which moved to `src/llm/provider/format-model-details.ts` so the CLI does not import from `src/tui/`. No match exits 1 with a single line, matching the honest-exit-code convention from #145.

## Verification

- `model-search.test.ts` — 12 cases: old substring behaviour preserved for one term, AND across terms, tag matching, vendor matching, rank ordering, stable ties, subsequence fallback, ids with no metadata.
- `models-search-command.test.ts` — 10 cases through the real command, including `--refresh` pulling a live catalog and every failure path exiting 1 with one line.
- `providers-inline-models-flow.test.ts` gains a case that presses `f` and types `qwen vision` through the **real key route** and asserts the list narrows to rows the catalog marks `supportsVision` — then that a third unsatisfiable term empties it.
- `npm run lint` clean; `npx vitest run src/cli src/llm src/tui/providers src/tui/llm-panel` green (657). Full `npm test`: 8 failures, all of which also fail on a clean `main` here (ink render smoke tests, a `$HOME` glob test, a sidecar timing test).

## Two things for review

1. `parseLlmProviderEntry` silently drops `userModels`, so that source can only be populated programmatically today even though the schema, the type and `resolveModel` all support it. Covered by a direct `collectHits` test rather than a config fixture, and left alone here — it wants its own fix.
2. The OpenRouter fetcher still filters Anthropic and Gemini out of the live catalog, so no query finds them. That is the subject of the companion PR; the two are independent.


sosidudku1:
The ranking is solid — deterministic and total via the `|| a.index - b.index` tiebreak so the list can't jitter, empty query short-circuits, and the query is never compiled into a RegExp, so `[`, `\` and `.*` are inert rather than throwing. CLI and TUI share one implementation.

**`model-search.ts:66` — the context-window tag is the display string, so `1m` matches almost nothing.**

```ts
tags.push(formatContextWindow(entry.contextWindow).toLowerCase());
```

`formatCompactNumber` returns `toFixed(1)` for non-integer millions, so 1,048,576 tags as `"1.0m"`, 1,310,720 as `"1.3m"`, 1,050,000 as `"1.1m"`. Tag matching is exact equality, and the AND semantics short-circuit on `RANK.none`, so a `1m` term eliminates the row.

The README advertises the query directly — `atomic-agent models search "1m cache" --provider openrouter` — and with #167's catalog that hides 19 of 54 chat rows: every Gemini, every GPT-5.x, both DeepSeek v4, both Llama 4, and several others.

What makes it worse than an empty result: `anthropic/claude-opus-5` is exactly 1,000,000, tags as `"1m"`, and does match. So the operator sees Claude rows, concludes Gemini and GPT-5 have no 1M-context variants, and moves on.

Cheapest fix is to push a normalized tag alongside the display string — both `1.0m` and `1m` — rather than changing the formatter and disturbing the rendered rows. Neither `model-search.test.ts` nor `models-search-command.test.ts` has a `1m` case.

plombeer31:
Fixed in 78e165c.

`formatContextWindow` is unchanged — the rendered rows are identical. `modelSearchTags` now pushes up to three deduped tags per window:

1. the display string (`1.0m`, `131k`), so searching for what a row shows still works;
2. the whole-unit **floor** in that same unit — 1_310_720 → `1m`, 1_050_000 → `1m`, 202_752 → `202k`. Floor rather than round, because a size term reads as a lower bound: `1m` should find everything from 1M up to 2M, and should not find a 950k row that would round up to it. So 1_310_720 answers to both `1m` and `1.3m`, as you suggested;
3. the **binary** reading when the window is an exact multiple of 1024 (1024² above a million) — 131_072 → `128k`, 262_144 → `256k`, 204_800 → `200k`, 1_048_576 → `1m`. Those windows are power-of-two sized and are advertised by the binary number; decimal rounding is what hid it. The exact-multiple guard keeps the reading off windows that were never binary, so 200_000 stays `200k` only.

Raw token counts (`131072`) are deliberately not tagged — no surface renders one.

Counts, chat rows only. Bundled OpenRouter catalog on this branch (18 rows): `1m` 3 → 9, `1m cache` 0 (exit 1) → 2, `256k` 0 → 3, `2m` 1 → 1. With #167 merged in locally (55 rows): `1m` 15 → 34, `1m cache` 13 → 30, `128k` 0 → 3, `200k` 1 → 3, `2m` unchanged at 1 — the 19 rows you counted are back.

Verified end to end with `models search "1m cache" --provider or`, which now returns the GPT-5.x rows instead of exiting 1.

Tests: `model-search.test.ts` gained a `context window terms` block (1_000_000 / 1_048_576 / 1_050_000 / 1_310_720 / 2_000_000 / 131_072 / 204_800, on digit-free ids so nothing matches via the id) plus a tag-shape case; `models-search-command.test.ts` gained a `1m`/`256k` case over the real bundled catalog. Four of the five new cases fail on the old implementation. `npm run lint` clean; the two files are 27/27, and the provider + TUI-providers suites (including `providers-inline-models-flow.test.ts`) are 255/255.

sosidudku1:
The `contextWindowTags` approach works — `1m` now returns every distinct 1M-class window (1_000_000, 1_048_576 and 1_050_000 all match) with no false positives, tags stay bounded at three per model and deduped, and the `b.score - a.score || a.index - b.index` tiebreak keeps ordering stable.

One documentation mismatch, no code change needed.

README describes a size term as "a lower bound whatever the row displays", but `contextWindowTags` floors to the whole unit — `contextWindowTags(2_000_000)` emits only `2m`, and tag matching is exact equality. So `1m` does not match a 2M window: against the bundled catalog, `openrouter/auto` and `x-ai/grok-4-fast-reasoning` (both 2_000_000) score 0 for `1m`. A user following the README's wording would expect the two largest-context models in the catalog to be in that result.

The source comment in `model-search.ts` states the actual intent precisely — "`1m` means 'a window of a million or better', so it must find every row from 1M up to 2M" — so the code is self-consistent and only the README overstates it. A one-line clarification there (e.g. "finds windows from 1M up to 2M") settles it.

plombeer31:
Treated the source comment as the spec and fixed the README — `73165c5`.

Both documents used the phrase "lower bound", but only one of them can be implemented as written: tags are matched by exact equality, which is what keeps the scorer a single lookup per term, and a genuine `>=` reading has to apply down the units too — `128k` would then match nearly the whole catalog and `1k` all of it, which is the opposite of what an ANDed narrowing term is for. Bucket semantics also keeps `1m` and `2m` disjoint rather than nested, so each term stays a filter instead of a prefix of a bigger result. So flooring is the behaviour worth keeping and the wording is what moved: the README now says a size term names a whole-unit bucket, `1m` finds windows from 1M up to 2M, a 2M window answers to `2m`. I tightened the source comment in the same direction — it led with "a window of a million or better", the same overstatement in miniature and plausibly where the README wording came from — and AGENTS.md, which carried "a lower bound" as well. Your reading of the catalog checks out on both bundled maps: `openrouter/auto` and `x-ai/grok-4-fast-reasoning` are 2_000_000, tagged `2m` alone.

On the test: the boundary was pinned only by `vendor/echo` being absent from a `toEqual` list, i.e. by omission, so it is now asserted by name (`searchModels(WINDOWS, "1m")` does not contain the 2M row) right where the `2m` case sits. `contextWindowTags` is untouched. Lint clean, 27/27 across the two search suites.

plombeer31:
Correction to my previous comment: `73165c5` landed on `valeryb/model-search` after this PR was already merged, so it never reached `main` — the README still carried the wrong wording. Carried over as #189 against current `main`, same content.

PR #169, opened 2026-08-19
feat(llm): drive Claude Code and Codex subscriptions through their own CLIs (no API key)

plombeer31:
## What

Adds a `subscription-cli` provider kind that runs an already-signed-in vendor CLI as an inference backend, so a flat-rate **Claude Code** or **OpenAI Codex** subscription can drive the agent with no API key and no per-token billing.

Every cloud kind we ship today (`openai-compatible`, `qwen-openai-compatible`, `openrouter`, `aimlapi`, `gemini`) requires a paid key. This adds the missing option for people who already pay for a subscription.

```jsonc
{
  "llm": {
    "activeTextProvider": "claude-cli",
    "providers": [
      { "id": "claude-cli", "kind": "subscription-cli",
        "defaultChatModel": "sonnet",
        "subscriptionCli": { "cli": "claude" } }
    ]
  }
}
```

In the TUI it is the first row of **Providers → `n`**, with no API-key screen.

## Design

One kind, parameterised by `subscriptionCli.cli`, with every CLI-specific byte behind a `CliAdapterDescriptor` (argv builders, output parser, hints). A third vendor CLI is a new descriptor, not a new provider kind — which keeps locked invariant #5 (plugin registration only, no central switches).

Three decisions worth reviewing:

**The prompt goes on stdin, never argv.** A full two-zone prompt exceeds the 128 KiB single-argument limit, so argv delivery would `E2BIG` on exactly the long sessions that matter most. Pinned by a test.

**`toolTransport` is `native_tools` even though this provider never returns `tool_calls`.** On the grammar transport, any format drift throws out of `parseToolCalls` and buys a second full CLI invocation on the repair path (~4 s, with no `maxTokens` to bound it, since neither CLI exposes that flag). On the native transport an empty `toolCalls` sends `step-executor` down its guarded recovery ladder, which parses the tool-call JSON out of `content` inside a try/catch and otherwise wraps prose as a `reply`. Same result when the model complies; no extra process when it doesn't.

**Disabling the child's own tools is safety-critical, not cosmetic.** Without it a second agent acts on the user's machine outside the approval ladder. `claude` gets `--tools ""` + `--strict-mcp-config`; `codex` gets `-s read-only` + `--ignore-user-config` — see the honest caveat below. `--bare` is never passed (its own docs say OAuth and keychain are never read under it, which would defeat the feature), and `ANTHROPIC_API_KEY` is neither set nor cleared.

New locked invariant **#8: subscription CLIs are driven, never impersonated** — documented headless mode only, no reading/extracting/replaying of OAuth tokens or keychain material. Pinned by argv tests.

## Verified against the real CLIs

Both legs were written against live output, not docs. Live agent turns, zero parse retries:

| CLI | Turn | Multi-step turn |
|---|---|---|
| `claude` 2.1.220 | `os.fs.read` → `reply` | `os.shell.run` → `os.fs.write` → `reply` |
| `codex` 0.148.0 | `os.fs.read` → `reply` | `os.fs.read` → `os.fs.write` → `reply` |

Probing found five things that would have been wrong if written from docs:

- `--output-format stream-json` **requires `--verbose`** under `--print`; it errors without it.
- `--json-schema` returns `stop_reason: "tool_use"` on success (it is a forced tool call underneath), so that maps to a normal stop.
- Claude's `modelUsage` includes the helper model Claude Code uses for its own side tasks — taking the first key reported `claude-haiku-4-5` as the model that served a `sonnet` request.
- **Codex exits 0 even when the turn fails.** A bad model id, an expired login and a rate limit all give a clean exit plus a `turn.failed` event, so the parser treats a missing `turn.completed` as failure rather than trusting the exit code.
- **Codex rejects every explicit model id under a ChatGPT login** (`not supported when using Codex with a ChatGPT account`) and resolves one server-side. I tested four ids mined from the binary; all four were rejected. So `-m` is omitted unless the operator sets one, and **no model list ships for Codex** rather than a guessed one.

## Known limitations (all documented in the README)

- **Codex has no `--tools ""` equivalent.** `-s read-only` confines its tools but cannot remove them. Left alone, Codex tries to *perform* the request with its own tools instead of emitting our protocol — the first live run answered "I can't find `probe.txt`" after looking in its own working directory. Since Codex also has no system-prompt flag, the steering is prepended to the prompt. It works, but it is a prompt-level guarantee, not a structural one like Claude's `--tools ""`.
- **No streaming on Codex** — `exec --json` emits one `item.completed` with no incremental text events, so `streamMode: "none"` and it buffers rather than pretending to stream.
- **~0.8 s of process startup per completion.** One inference per step, so a 10-step turn pays ~8 s of spawn overhead. Server-side prompt caching does survive across separate invocations (verified `cache_read_input_tokens` on a fresh process), so the KV-stable prompt is not wasted — the cost is spawn, not reprocessing.
- **Subscription rate limits** are hit far faster by an autonomous agent than by interactive use. There is no structured signal, so the CLI's own message is surfaced verbatim rather than summarised away. Claude's stream also carries a `rate_limit_event`, which is logged as a warning.
- **Sampling knobs are inert** (`temperature`, `top_p`, `top_k`, `seed`, `stop`, `maxTokens`) — neither CLI exposes a flag, so they are dropped rather than silently approximated. `REPAIR_MAX_TOKENS` is the concrete casualty, which is another reason for the native transport.
- **No vision, no embeddings** — embeddings stay on the local daemon, which is already the default.
- Windows `.cmd` resolution is implemented but is the least-tested path.

**ToS posture:** we invoke the officially documented headless mode and let each CLI authenticate itself. Whether driving a subscription CLI from another agent is acceptable use is the vendor's call; the README says so and leaves the decision to the operator.

## Deliberately not in this PR

Reconfigure-in-wizard (`isCloudProviderKind` stays unextended, so configure-mode is never entered with a kind it cannot represent — the model is still changeable from the Cloud pane), cost display for subscription turns, and deriving a real JSON Schema for the tool-call batch to enforce via `--json-schema` (the principled GBNF equivalent, and a PR of its own).

## Notes for review

- `USER_CONFIG_VERSION` stays **37**. `parseUserConfigFile` gates only on `SUPPORTED_INPUT_VERSIONS` and fills absent blocks from defaults, so additive optional fields need no migration. A bump would be actively harmful: released builds top out at 37 and would reject the *entire* config file rather than just an unknown entry.
- `parseLlmProviderEntry` is a whitelist that rebuilds the entry from known keys, so `subscriptionCli` had to be parsed explicitly or it would be silently dropped on the next config rewrite. Pinned by a round-trip test.
- The startup gate gains one branch, reachable only for a kind that config validation rejected outright before this PR — no existing config changes behaviour.
- Two wizard row-order tests were rewritten to address rows by name (`KIND_ROW_ORDER.indexOf(...)`) instead of by keystroke count, so inserting a row no longer silently repoints them at a different provider.

## Testing

`npm run lint` and `npm run build` are green. 70 new tests; no test spawns a vendor CLI (the runner is injected, mirroring `OpenAiProviderOptions.fetchImpl`).

Full-suite failures are unchanged from `main`: 4 pre-existing failures (`ChatLog`, `SplashBanner`, two `TuiApp` smoke tests) reproduce on a clean checkout, and `llm-health-poller` / `persistEmbeddingHybridRecall` / the `os.git.*` suite fail only under parallel load and pass in isolation.


sosidudku1:
The security-critical parts are done right, and I want to say that first: the prompt goes on the child's stdin rather than argv, both spawn sites pass an argv array with `shell: false`, and the 128 KiB ARG_MAX limit is called out in the code as the reason for that choice. Binary resolution avoids a shell PATH lookup, and `JSON.parse` is guarded on both adapters.

Two blockers, both in process lifecycle, both the same root cause.

**1. An unhandled `EPIPE` on stdin kills the whole agent.** `command-runner.ts:154` writes to `child.stdin` with no `error` listener anywhere in the file. When the CLI exits early — signed out, wrong model, rate-limited — it does not drain stdin. A prompt over ~64 KiB cannot flush and emits `EPIPE` on a stream with no handler, and `installGlobalErrorHandlers` deliberately preserves Node's fatal semantics, so this reaches `process.exit(1)`.

That means the user loses the entire session instead of seeing the `SubscriptionCliAuthError` this PR builds for exactly that case. Reproduced: 16 KiB and 64 KiB complete cleanly; 128 KiB, 256 KiB and 1 MiB all produce `UNCAUGHT EXCEPTION: EPIPE`. Worth noting the prompt size is not exotic — the comment at `subscription-cli-provider.ts:271` says a two-zone prompt "routinely exceeds the 128 KiB single-argument limit".

**2. The same write in `stream-cli-completion.ts:95`, where our own abort is the trigger.** Ctrl+C fires `onAbort` → `stop("abort")` → SIGTERM to the child while the stdin payload is still draining → same EPIPE → same fatal exit. Cancelling a turn is the most routine action in the TUI.

This defect was latent in `command-runner.ts` before this PR, but unreachable — the existing git and test-runner callers stay well under the pipe buffer. This PR is what makes it reachable.

Fix for both: an `error` handler on `child.stdin` (EPIPE is expected here, not exceptional).

**3. `stream-cli-completion.ts:138` — the `finally` cancels the SIGKILL escalation it just scheduled.**

```ts
if (!settled) stop("done");            // may set killTimer
if (killTimer) clearTimeout(killTimer); // always clears it
```

The clear runs milliseconds later, long before the 2s delay, so a child that traps SIGTERM is never force-killed. Reproduced with a `trap '' TERM` child: the process survives as an orphan, one per aborted turn. Only clear `killTimer` when the child actually exited.

**4. Minor:** `isCloudProviderKind` wasn't extended with the new kind, so pressing `c` on a `claude-cli` row in the Providers panel consumes the key and does nothing — no wizard, no error.

The grammar handling is honest and I'd leave it as-is: declaring `native_tools` while never returning `tool_calls` routes into step-executor's guarded recovery ladder, which I checked does not throw on malformed output.

plombeer31:
All four addressed. Three commits, one per finding area: 9b58547, 61bf400, 517ca89.

**1 + 2 — EPIPE on stdin (`command-runner.ts:154`, `stream-cli-completion.ts:95`).** Reproduced your boundary exactly against a child that exits without reading: 16 KiB and 64 KiB land in the pipe buffer and complete cleanly, 128 KiB / 256 KiB / 1 MiB all give `UNCAUGHT EXCEPTION: EPIPE`. Both sites now carry an `error` listener on `child.stdin`.

It absorbs broken-pipe codes only (`EPIPE`, `ECONNRESET`, `EOF`, `ERR_STREAM_DESTROYED`, `ERR_STREAM_WRITE_AFTER_END`); every other stream error still travels — `runCommand` rejects with it, the streaming generator throws it after the stdout drain. The signed-out case now arrives as `SubscriptionCliAuthError` because `close` still carries the exit code and stderr into `mapCliFailure`.

One thing the plain absorb would have made silent: `codex` exits 0 even on failure, so a run that dropped the pipe halfway through the prompt would have looked like a clean completion over a prompt we never finished sending. Added `CommandResult.inputTruncated`, which is its own failure condition in both runners and maps to its own message, checked after the auth patterns so "run /login" still wins.

**3 — the `finally` cancelling its own SIGKILL escalation.** Reproduced with a child that ignores SIGTERM: before the fix it outlives an 8s watch as an orphan, after it dies inside the 2s escalation window. `killTimer` is now cleared only once the child has actually exited (disarmed via the exit promise otherwise), and `stop` no longer re-arms an escalation that is already pending — abort followed by the generator's own cleanup used to orphan the first timer.

**4 — `isCloudProviderKind`.** Extending that predicate is not quite the fix: two wizard rows (`claude-cli`, `codex-cli`) collapse onto one config kind, so a row's stored `kind` is `subscription-cli` and cannot say which CLI it drives — only `entry.subscriptionCli.cli` can. Added `configureWizardKindForRow` as the single "can this row be configured, and as what" answer, used by `c` and by the LLM tab's configure action; `isCloudProviderKind` stays what its name claims.

On what `c` should do there: the cloud wizard as it stood is the wrong destination, for your reason. A CLI-backed provider has no key to paste and no endpoint to type, so configure's `api_key` opening screen is the same dead end `advanceWizardPhase` already skips on the add path. It now opens on `chat_model_line` — the one editable thing — prefilled with the pinned model so Enter keeps it rather than resetting it to the adapter default, with the CLI's own placeholder (`sonnet` for claude, "the CLI resolves the model" for codex, which rejects explicit ids under a ChatGPT login) instead of `gpt-5.4-mini`, and no model-list fetch, since there is no endpoint behind it. Refusing the key with a message would have been the alternative, but the model id is genuinely configurable and there was no other way to reach it from that panel.

Grammar handling left alone as you said.

Tests: new `src/sandbox/command-runner.test.ts`, `run-cli-completion.test.ts` and `providers-key-bindings.test.ts`, plus additions to `stream-cli-completion.test.ts`, `providers-wizard.test.tsx`, `providers-orchestrator.test.ts` and `providers-wizard-key-bindings.test.ts`. The process-level ones spawn real children (a mock cannot reproduce what the kernel does to a pipe with no reader); the orphan test reads the child's own pid off its first line and force-kills anything left in `afterEach`. Each was checked against a revert of its source hunk: reverting the stdin listener turns the suites red with `Uncaught Exception: write EPIPE`, reverting the `killTimer` change leaves the trapping child alive for the full watch, reverting the row-kind resolution or the phase choice fails the panel and render tests.

`npm run lint` clean. Targeted run (`subscription-cli/`, `sandbox/`, `tui/providers/`, `tui/llm-panel/`, `providers-wizard.test.tsx`, the two `CommandResult` helper users): 32 files, 354 tests, all passing. Full suite: 6 files / 8 tests failing, the same set as stock HEAD (splash-banner fixture — which `chat-log` shares, `tui-app` fixtures ×3, `localModels.embeddings.url`, the dev-machine `fs-glob` path, `send-message-concurrency`); confirmed at `3d35833` before blaming the diff.

AGENTS.md gains invariant 9 under the provider section, so the stdin listener and the escalation ordering are not quietly reverted later.

PR #172, opened 2026-08-19
feat(tui): Local · Cloud · Fusion in the ctrl+p menu, with ctrl+g 1/2/3 to pick one

plombeer31:
Closes the gap you flagged: the run modes existed (#161 → #162 → #163) but the operator menu did not know about them, so there was no way to reach them by browsing and no direct key to pick one.

> **Stacking.** This is a diamond: it needs both #163 (the run modes) and #170 → #171 (the menu) as ancestors, and a PR can only have one base. Base is `feat/run-mode-tui`, so GitHub shows the menu stack in the diff as well. **The reviewable delta is the single `integrate:` commit** — everything before it is #170 and #171, already reviewed there.

## What you get

```
RUN
  Run type…                                     
  Local                  Run type      ctrl+g 1
  Cloud                  Run type      ctrl+g 2
  Fusion                 Run type      ctrl+g 3
```

- **Browsable** under `Run type ▸`, one level deep like Observe and Manage.
- **Searchable** — typing `run` in the menu surfaces all three with a `Run type` breadcrumb.
- **Live state** — the `Run type` row shows the mode in force (`fusion · 40% cloud`), so the menu answers "what am I on" as well as "what can I switch to".
- **Direct keys** — `ctrl+g 1/2/3`. Digits were unclaimed in the leader namespace and read naturally as "pick option N". `ctrl+r` (cycle) and `/run` (dial picker) from #163 are untouched.

## How it is wired

The registry gains a second activation channel. `slash` means *this node is listed in the palette and activated by running `/name`*; the new `command` means *activated by running this literal line*, and is not listed:

```ts
{ kind: "action", id: "run.mode.fusion", label: "Fusion",
  group: "run", parent: "run.mode", chord: "3", command: "/run fusion" },
```

So the palette still lists `/run` once rather than once per mode, and the menu still has exactly one dispatch path — `slash-command-handler.ts`. Nothing in this PR reimplements run-mode logic; every entry goes through #163's own command, including its degradation rules.

## Conflicts resolved

Two, both in files the menu stack had already touched:

- **`slash-commands.ts`** — kept the derived form from #170; `/run` now arrives through the registry, and `chat` loses its `run` alias to match #163.
- **`hotkey-hint.tsx`** — kept the `ctrl+p menu` chip and #163's note that `ctrl+r` stays unadvertised because the mode strip above the chat is its visible entry point.

## The palette test is now a cross-check, not a snapshot

#170's test pinned the derived palette against a v0.2.2 snapshot. Here the palette legitimately changes, so rather than blessing my own output I extracted #163's hand-written `SLASH_COMMANDS` and made *that* the expected value:

```ts
it("derives exactly the palette #163 declares — same commands, same order")
```

If the registry and #163 ever disagree about a command, its description, its aliases or its position, this fails. It passes exactly today.

## Verification

- `npm run lint` clean.
- `npx vitest run src/tui` — 1041 passing; the same pre-existing failures as `main`, none new. #163's own run-mode suites (`run-mode-nav`, `run-mode-reducer`, `run-mode-key-bindings`) all pass unchanged.
- Driven in a real PTY at 100×30: searching `run` lists all three modes with their chords, and `ctrl+g 3` with no cloud provider configured produces #163's own degradation message — *"Fusion needs a cloud orchestrator — no cloud provider is configured. Staying on local."* — which is the proof the chord dispatches through the real path rather than a parallel one.

### DIFF
--- a/src/tui/app-key-bindings.ts
+++ b/src/tui/app-key-bindings.ts
@@ -6,6 +6,13 @@ import {
   type ApprovalRequest,
 } from "../approval/approval-gate.js";
 import { formatApprovalCategory } from "../approval/approval-level.js";
+import {
+  handleMenuKey,
+  isMenuLeaderKey,
+  isMenuOpenKey,
+  resolveLeaderChord,
+} from "./menu/menu-keys.js";
+import type { MenuNode } from "./menu/menu-registry.js";
 import { cycleNavSlot, type NavSlot } from "./section.js";
 import { selectSidebarTasks } from "./sidebar-tasks-selector.js";
 import type { TuiAction } from "./tui-action.js";
@@ -72,6 +79,11 @@ export interface AppKeyContext {
    * the sidebar steals plain Tab.
    */
   sidebarVisible: boolean;
+  /** True while a `ctrl+g` leader is waiting for its chord key. */
+  menuLeaderArmed: boolean;
+  setMenuLeaderArmed: (armed: boolean) => void;
+  /** Navigate to a place, or run an action's slash command. */
+  activateMenuNode: (node: MenuNode) => void;
 }
 
 /**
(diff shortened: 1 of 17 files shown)


plombeer31:

PR #173, opened 2026-08-19
feat(tui): every visible control takes a click — menu rows, run-mode pills, the dial

plombeer31:
> **Stacked on #172**, and merges #165 (`feat/tui-mouse`) so the two can meet. Review #172 first; the commits that belong to *this* PR are the last two.

## Why this branch exists

Hand testing turned up "mouse clicks need to work in the menu and in the popup nav menu". No single existing PR can carry that fix: the mouse layer is #165, the menu is #170/#171, the run modes are #161–#163, and none of those branches contains the others. The surfaces only exist together here — the same reason #172 exists.

## The one bug behind three symptoms

`TuiApp` raises the mouse registry's layer floor whenever a modal opens, so a click can never reach the list rendered behind it. Neither the `ctrl+p` menu nor the run-mode picker registered its own targets at `MOUSE_LAYER_MODAL`, so both rendered perfectly and swallowed every click that landed on them. Everything else here is wiring on top of that.

## What became clickable

| Surface | Gesture |
|---|---|
| Menu rows | One click acts. A submenu row opens the submenu, same as `→` |
| Run-mode pills (`Local · Cloud · Fusion`) | Click switches mode |
| The pill already in effect | Opens the dial instead of re-applying the mode |
| The cloud-share dial | It is a slider — clicking column N sets the share that column represents |
| Run-mode picker rows | Two-step: click selects, second click applies |
| The breadcrumb | Opens the menu |

**Why the menu is the exception to the two-step rule.** Everywhere else in the mouse layer, the first click selects and the second activates, because a mis-click there starts a model download or switches sessions. In a menu the operator has opened it specifically to pick something, so making them click twice is the surprising choice.

**Why the breadcrumb is clickable.** #172 retired the Run / Observe / Manage pills in favour of a breadcrumb, which removed the only mouse route into navigation. A click on it opens the menu, so "everything you can see, you can click" still holds.

## Tests

- Five of #165's mouse tests waited on the `▸ Run` pill #172 removed; they go through the breadcrumb now.
- New: clicking a menu row navigates; clicking a pill asks for that run mode; clicking the active pill opens the dial.
- `npm run lint` clean. Full suite: 4288 passed / 5 failed — three are environment-dependent on this machine (`fs-glob-real` looks for a CV in `$HOME`, `send-message-concurrency` and `persist-embedding-hybrid-recall` flake under parallel load) and two (`chat-log`, `splash-banner`) are the pre-existing stale-copy failures #170's body already lists for `main`. No new failures.

## Verified in a real terminal

Driven through a PTY against the built `dist/` at 120×40, 100×24, 80×30 and 60×20, rendered through a terminal emulator: the menu is centred on both axes at every size, a click on a menu row navigates, a click on a pill switches mode, and a click on the active pill opens the dial.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -147,6 +147,22 @@ Speculative batching (the runtime guessing that the model "should" have batched 
 - Tests are colocated with source: `build-prompt.test.ts` next to `build-prompt.ts`.
 - Config lives in `src/config/` — read it before touching env vars.
 
+## Mouse support
+
+The TUI is clickable. Ink has no mouse layer, so this is built in `src/tui/mouse/`:
+
+1. **Reporting** — `enableMouseTracking` writes `\x1b[?1000h\x1b[?1006h` (button events + SGR coordinates). 1002/1003 motion tracking is deliberately **not** requested: nothing in the UI hovers or drags, and motion reports are a constant wakeup stream. Paired with a `process.on("exit")` restore, like `alt-screen.ts`.
+2. **Decoding** — `decodeMouseEvents` is a pure function over a stdin chunk returning `{ events, text, rest }`. It understands SGR and legacy X10, buffers a report split across two reads, and passes a lone trailing `ESC` straight through (buffering it would delay the Escape key by one keystroke).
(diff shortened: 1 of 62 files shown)


plombeer31:

PR #175, opened 2026-08-19
atomic-agent 0.3.0 — every open PR integrated, plus what four rounds of hand testing found

plombeer31:
The 0.3.0 integration build: every open PR of mine merged into one branch, plus the fixes that came out of testing the result by hand.

## What is in it

**All 24 open PRs** — #150–#173 — merged in dependency order. Nothing from other contributors.

**Then four rounds of hand testing**, each of which turned up things no individual PR could have caught, because they only exist where two PRs meet.

## The bugs that only appeared once everything was together

**Config version skew made the release unusable.** #161 and #165 both bump `USER_CONFIG_VERSION` to 38. Two builds share one `~/.atomic-agent/config.json`, so the moment this one wrote v38 the installed v0.2.2 died on *every* command — `models status`, `config get`, the TUI — with `unsupported config version 38; expected one of 5, 6, … 37`. The supported-version list was an enumeration with no upper bound. Newer versions are now read rather than refused, and `ensureUserConfigFileSync` never rewrites a newer file down (with a shared file, two builds would otherwise take turns destroying each other's keys).

**Asset directories resolved against the working directory.** `resolveAssetDir` looked next to the Node binary, then fell back to `cwd`. Starting the agent by absolute path from an unrelated directory — exactly what #150's Ctrl+N spawn generates — died with `ENOENT … /grammars/tool-call.gbnf`. It now tries a module-relative path in between. Ctrl+N also forwards the asset-dir env vars, since a spawned login shell inherits nothing.

**Enter in the `/run` overlay did nothing.** #163 dispatched `run_mode_change_requested`; the reducer no-ops it by design and the orchestrator never saw it, because the bus is bridged into the reducer one way. The *mouse* path already used the callback — so clicking a selected row applied the mode and pressing Enter on it did not. The old test asserted the dispatch, so it passed for the whole life of the bug.

**The add-provider wizard painted over its own rows.** It drew at the top of `LlmPanel` with the route card, mode header and row list still rendered underneath, and the panel already spends its whole row budget on those. The frame ran ~16 rows past the terminal, and Ink 7 paints later lines over earlier ones instead of clipping — which is why OpenRouter was invisible at some window heights and the list looked like it held one provider.

**A refused API key was invisible.** #166's verification worked — the key was rejected and nothing was written to disk — but the list screens had no error line and no busy state, so Enter ran the check, swallowed every key but Esc, and repainted an identical screen. It read as being stuck on model selection.

**The tasks table was laid out for a width it never had.** Rows padded fixed columns totalling ~123 while the panel had 73–88, so every row wrapped and the cells collided; and the table drew ~5 rows more than the pane budgeted, overpainting the filter bar and column header. The key surface was fine the whole time — the hint strip explaining it was being overpainted too.

## The UI round

The chrome was reorganised after looking at the running app:

- The **left rail** replaces the top bar: brand mark, version, session id, a Menu button, Sessions and Tasks, on its own inverted ground (per-palette, never a literal white — `#fff` disappears on the four light themes). Below the rail's width threshold the one-row bar returns, so a narrow terminal is not left with no chrome at all.
- The **menu** is a centred modal that dims the app behind it, takes clicks on its rows, scrolls with the wheel, and closes on a click anywhere outside. `Esc` on an idle Run screen opens it.
- The **composer** is a framed field with a Send button on the same inverted ground.
- Every message carries **`[copy]`** and, on user messages, **`[try again]`**.
- The **run-mode strip** moved below the composer, says how to change it, switches the composer's model immediately, and routes an unconfigured mode to the screen that configures it.
- The **brand mark** is scaled from one drawing by `logo-raster.ts` rather than hand-drawn per size, so every size is the same silhouette.

## Two honest notes

**Mouse text selection is not implemented in-app**, deliberately. It would need motion tracking (off on purpose), a way to read which character is painted in each cell (Ink exposes sizes only), and every component under the selection rectangle to become selection-aware — and the result would still be worse than the terminal's own, copying rendered text with borders and wrap points baked in. Shift+drag already works wherever the terminal supports it; Apple Terminal gets a 10-second reporting pause instead. That pause is **unverified against Terminal.app itself** from this environment.

**"Arrows don't work the first time you open `/model`" could not be reproduced.** An adversarial verifier drove the unmodified build at two sizes, six repeats, including with a deliberately slow catalog, and the first arrow moved the cursor every time. The reducer change that shipped is sound and covered, but it targets a window that the app's own wiring does not appear to reach. If it reproduces for you, the exact provider configuration matters and I would like it.

## Verification

`npm run lint` clean. **4816 passed, 3 failed** — `fs-glob-real` (looks for a CV in `$HOME`), `send-message-concurrency` and `persist-embedding-hybrid-recall`, all three failing identically on pristine `main` in this environment.

Every UI change was driven through a PTY against the built `dist/` at 120×40, 100×30, 80×24 and 60×20 and read back through a terminal emulator, not asserted from the source. The four fixes in the last round each went through an independent verifier whose brief was to refute them, including reverting the source and re-running the tests to prove they fail without the fix, and mutation-testing the guards.


sosidudku1:
Reviewed only the work unique to this branch (after `3e6ba4a`) — the individual PRs were reviewed separately, so nothing here re-litigates those.

Two things worth fixing before this ships.

**1. `openai-stream-consumer.ts:155` — the de-duplication truncates real tool names.**

```ts
if (fn.name === fragment || fn.name.endsWith(fragment)) return;
```

The second clause drops a genuine fragment whenever it happens to be a suffix of what's already accumulated. For a gateway that splits names across chunks — the case the third new test explicitly commits to supporting — `"os.proc.kil"` + `"l"` stays `os.proc.kil`, misses the registry, and fails the turn with exactly the "tool not registered" error this commit set out to fix.

Enumerating two-way splits of the repo's real tool names, eight are lossy: `os.proc.kill`, `os.fs.diff`, `os.git.diff`, `browser.scroll`, the three `memory.*.recall` tools, and `app` — every name ending in a doubled letter, split before the final character.

The signal can't separate a repeat from a continuation even in principle (`["search","search"]` for `searchsearch` yields `search`). The first clause already covers the reported shape, where the whole name repeats each chunk; the `endsWith` clause only adds loss. The four new tests pass because none exercises a lossy split — the relevant one uses `os.fs` + `.read`, which happens to be safe.

**2. `config-schema.ts:2704` — the comment claims a property the code doesn't have.**

The justification for permissive newer-version reading says `writeUserConfigFileSync` "preserves unknown top-level keys rather than dropping them." It doesn't — it's `JSON.stringify` over a `UserConfigFile` that `parseUserConfigFile` rebuilds field-by-field from a fixed key list, with no passthrough.

So a newer build's config survives being read (the fix works, and `ensureUserConfigFileSync` correctly declines to rewrite), but the first `config set` — or any of the other persist callers — drops the unknown key and stamps `version` back down. The loop this is meant to prevent is delayed until the operator's first setting change, not prevented.

The permissiveness itself is defensible: the version floor is retained and the integer check is a real tightening. It's the stated premise that needs correcting, either in the comment or by making the write actually preserve unknown keys.

**Checked and clean:** the asset-dir fix is correct for both SEA and global npm installs (the `execPath` probe runs before the new module-relative step, and `cwd` correctly stays last). I looked specifically for merge-resolution damage in the key-binding dispatch, where several PRs collided, and found none — the Esc ladder preserves every rung in order, and the new menu rung re-checks all prior claimants rather than trusting fallthrough. Test delta against `main` is zero; the one failure at HEAD reproduces identically on `main`.

plombeer31:
Both findings fixed on `release/v0.3.0`.

**1. Tool-name truncation** (`f23e8a6`) — the `endsWith` clause is gone; only an exact whole-name repeat is dropped now. You're right that the signal can't tell a partial repeat from a continuation, and it only ever loses characters: `os.proc.kil` + `l` was landing as `os.proc.kil` and failing the same registry lookup the commit set out to fix. The doc comment above `appendToolName` no longer claims the two cases are separable — it states that only a whole-name repeat is detectable, with the doubled-letter names as the reason not to guess. New test pins `os.proc.kil` + `l` → `os.proc.kill`.

**2. `config-schema.ts` comment vs. code** (`faabd9a`) — made the code match the comment, per the owner's call.

- `parseUserConfigFile` collects the top-level keys it has no parser for and carries them on the returned object under a new `UNKNOWN_USER_CONFIG_KEYS` symbol. Symbol, not a field: no JSON key can collide with it, `JSON.stringify` skips it (so a preserved key is written once, by the merge, never twice), and the `{ ...prev, tui: … }` drafts in every `persist-*` helper carry it for free. The collector unions literal keys with the carrier, so the read → merge → validate → write round trip doesn't drop them on the second validation. The known-key set derives from `USER_CONFIG_DEFAULTS` plus the two input-only keys (`llm`, legacy `telemetry`), so a new block registers itself.
- `writeUserConfigFileSync` merges the carrier **and** the unknown keys still in the file on disk back into the payload, and leaves a newer on-disk `version` standing rather than stamping it down. Both sources are needed: the carrier covers the `persist-*` helpers; the disk read covers the whole-payload replacements (`config set`, `PATCH /api/config`) whose payload never saw those keys. Keys this build owns always win, so a preserved key can't shadow a parsed value.

Tests: carrier survives parse → spread → parse and is absent when the file is fully understood; on disk, a v43 file with a foreign top-level key keeps both the key and its version through a `persist-*` rewrite and through a whole-file replacement.

Verification: `npm run lint` clean. Full suite before my changes on this branch: 6 files / 7 tests failing (llm-health-poller ×2, send-message-concurrency, persist-embedding-hybrid-recall, fs-glob-real, and two `os.git.*` timeouts). After: same set modulo the `os.git.*` family, which times out non-deterministically on this machine — re-running `src/tools/os/git/` on the untouched HEAD gives 12 failures vs 11 with the changes applied, so it's environmental. `src/config/**`, `src/llm/provider/openai/**`, `config-command`, `route-config`, `persist-user-tui-config` and `telegram-settings`: 20 files / 289 tests, all green. Net new tests: 5.

Known limitation, not addressed: only *top-level* keys are preserved. A new sub-key inside a block this build knows (e.g. `memory.somethingNew`) is still dropped by the field-by-field parse.

sosidudku1:
Closing this one, and I want to explain the reasoning properly, because the work in here was not wasted.

## What happened since this branch was cut

`release/v0.3.0` branches from `667dae1` (v0.2.2). Since then main has taken 25 PRs, including your whole non-fusion stack, reviewed and landed individually:

- **Esc / editor:** #155, #156, #158 — these turned out to be three incompatible iterations of one design, so they were merged as a single priority ladder rather than literally; a verbatim merge would have double-aborted on every Esc.
- **Steering:** #157, #159, #160 — with three "accepted implies delivered" gaps closed in the sidecar path.
- **Independent:** #150 (ctrl+n), #169 (Claude Code / Codex via their own CLIs).
- Plus #176, #177, #187, #188, #189 and the GAIA reliability work in #191.

## Why this branch can no longer be merged as-is

It is not that the diff is stale — it is that it now moves backwards. Measured against current main rather than the old merge base:

```
git diff origin/main..pr175 --shortstat
277 files changed, 14298 insertions(+), 9267 deletions(-)
```

Those 9267 deletions are real. None of today's commits are ancestors of this branch, so merging it would remove 15 files that landed today, among them:

- `src/llm/provider/openai/ascii-header-guard.ts` (+ test) — the non-ASCII key fix from #187
- `src/http/undelivered-steers.ts` (+ test) — the delivery-guarantee fix from #160
- `src/llm/provider/subscription-cli/run-cli-completion.test.ts` — #169
- `src/tools/coerce-tool-args.ts` (+ test) — #191

The config-version line tells the same story: main is on `USER_CONFIG_VERSION = 39`, this branch is on 38.

Reconciling that would mean re-merging 25 PRs into a 286-file branch and re-testing the result — strictly more work than carrying the two things that are genuinely still missing, on top of a main that is already green.

## What is actually still outstanding, and where it goes

Only two areas from this integration are not in main:

1. **The fusion / run-mode stack** — #161 → #162 → #163 → #172, with #173 sitting on top of them. Untouched by today's merges and still the real review unit. That review is a separate decision, not a casualty of this close.
2. **The mouse layer** (#165) — being ported onto current main directly, as its own PR.

## The findings from your hand-testing rounds

These were the most valuable part of this PR and they should not disappear with it, since several are cross-cutting bugs no single branch could surface:

- the supported-config-version list being a closed enumeration, so a newer build bricks an older one sharing `~/.atomic-agent/config.json`
- `resolveAssetDir` falling back to `cwd`, which breaks precisely the ctrl+n spawn from #150
- Enter in the `/run` overlay dispatching an action the orchestrator never sees, while the mouse path used the callback and worked
- the add-provider wizard and the tasks table both overrunning their row budget, with Ink 7 overpainting instead of clipping

The first two apply to main today and are not fixed there. They are worth one small PR each against current main, where they can be reviewed on their own merits and land in a day instead of waiting on a 286-file branch. Happy to pick them up if you would rather not re-cut them.

Thanks for doing the integration and the four rounds of manual testing — the per-PR reviews were substantially faster because this branch had already found where the seams were.

PR #190, opened 2026-08-20
test(tui): follow the breadcrumb and the block splash mark

sosidudku1:
Seven tests were failing on `main` against code that behaves correctly. They assert on a UI surface that two merged PRs replaced.

**What changed under them**

- **#171** replaced the three-section pill row (`▸ Observe`) with a breadcrumb, so the screen now reads `Observe ▸ Feed`. The section label is the left half of a crumb rather than a standalone pill.
- **#151** made the splash mark a scaled block raster (`█`), so the `:::` glyphs the smoke test looked for no longer appear at any size.

**Why this is a test fix and not a code fix**

The behavioural assertion in the same file — `does not quit when Esc is pressed twice from an Observe tab`, which counts `quit`/`abort` callbacks rather than matching text — passed throughout. Esc has been returning to Run correctly the whole time; only the strings moved.

**The changes**

- `escape-observe-tabs.test.tsx` — `▸ Observe` → `Observe ▸`, `▸ Run` → `Run`
- `escape-import-tab.test.tsx` — `▸ Manage` → `Manage ▸`, `▸ Run` → `Run`
- `tui-app.test.tsx` — the splash assertion now checks the rail the Run screen always carries, since the mark's glyphs vary with the surface

Tests only; no production code is touched.

**Delta**

`v0.2.2` baseline: 7 failures / 4075 tests. Before this PR: 14 / 4309. After: 5 / 4309.

The five that remain were all failing on `v0.2.2` too and are out of scope here: two `TuiApp (smoke)` cases (`Shift+Tab` expects Telegram as the last Manage tab, now Privacy; `two-mode LLM panel` expects a heading the panel no longer renders), plus `persistEmbeddingHybridRecall`, the `Sibiliainen CV` search case, and the sidecar FIFO concurrency test. `Esc from an idle Manage panel` also appears intermittently — it passes in isolation and settles ~10ms after the write, under Ink's 20ms flush delay.

### DIFF
--- a/src/tui/escape-import-tab.test.tsx
+++ b/src/tui/escape-import-tab.test.tsx
@@ -49,7 +49,7 @@ describe("Esc on the Import tab", () => {
     bus.emit({ type: "ui_mode_set", mode: "debug" });
     bus.emit({ type: "tab_changed", tab: "import" });
     await settle();
-    expect(strip(lastFrame() ?? "")).toContain("▸ Manage");
+    expect(strip(lastFrame() ?? "")).toContain("Manage ▸");
 
     stdin.write(ESC);
     await settle();
@@ -57,7 +57,7 @@ describe("Esc on the Import tab", () => {
     // The configure-mode handler ends in a catch-all `return true` that
     // swallows stray letters; before the fix it swallowed Esc too, so the
     // operator was stuck on the tab with no "back" gesture at all.
-    expect(strip(lastFrame() ?? "")).toContain("▸ Run");
+    expect(strip(lastFrame() ?? "")).toContain("Run");
     expect(quit).toBe(0);
     unmount();
   });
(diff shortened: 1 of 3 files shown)


sosidudku1:
Pushed two more fixes so this closes out the stale-expectation cleanup completely rather than leaving two behind.

- **Shift+Tab from Run** expected Telegram as the last Manage sub-tab; `MANAGE_TABS` gained `import` and `privacy` since, so the wrap target is Privacy. The assertion carries the `▸` active marker — a bare "Privacy" also matches the inactive chip in the strip, which made the first version of this fix pass even with `privacy` removed from the list.
- **The two-mode LLM panel** asserted `RouteCard` chrome that correct code deliberately drops: ink-testing-library reports no rows, so `LlmPanel` falls back to 80x24 and picks its compact header. The full/compact decision is already owned by `components/llm-panel.test.tsx`, which drives `maxRows` explicitly, so the smoke test now asserts the two-mode body every budget keeps.

Both were checked by ablation — each fails when the behaviour it describes is removed.

Suite on this branch over current main (which now includes #165): **4766 tests, 2 failures**, down from 5. The two left are the flaky sidecar FIFO case and the `$HOME`-dependent glob test, both failing on pristine main and out of scope here.

PR #207, opened 2026-08-20
fix(agent): derive parallel_tool_calls from cap and provider capability

### LINKED ISSUE #104: maxParallelToolCalls=1 still sends parallel_tool_calls: true to OpenAI-compatible providers

yablokolabs:
## Summary

Closes #104. `buildLlmStreamParams` hardcoded `parallelToolCalls: true`, so setting `agent.maxParallelToolCalls = 1` could never act as a provider-compatibility control — the request body still advertised parallel tool calls to backends that reject or mishandle them (e.g. Gemini). The flag now derives from **both**:

- the configured cap: `agent.maxParallelToolCalls > 1`, and
- the active provider's `supportsParallelTools` capability.

## Implementation

- `AgentLoopDeps` / `StepDependencies` gain `supportsParallelTools`, threaded from the resolved provider slice (`resolveActiveLlmSlice().provider.capabilities`) in `bootstrap.ts`.
- `buildLlmStreamParams` computes `parallelToolCalls: maxParallelToolCalls > 1 && supportsParallelTools`.
- `openai-build-body` propagates it; non-tool requests still never gain the field.

## Tests

- `step-executor.test.ts`: 4 new cases — cap=1 ⇒ `false`, provider lacking support ⇒ `false`, cap>1 + support ⇒ `true`, exact-cap batch unchanged.
- `openai-build-body.test.ts`: 2 new cases pinning `false` and absent-field propagation at the body level.
- Lint clean; full `src/agent` + `src/llm/provider/openai` suites green.

### DIFF
--- a/src/agent/agent-loop.ts
+++ b/src/agent/agent-loop.ts
@@ -77,6 +77,13 @@ export interface AgentLoopDependencies {
   toolCallAdapter?: ToolCallAdapter | null;
   supportsSlotAffinity?: boolean;
   /**
+   * Whether the active native-tools provider can emit parallel tool
+   * calls. Defaults to `true` when omitted (legacy / grammar-only
+   * wiring). Combined with `agent.maxParallelToolCalls` to decide the
+   * `parallel_tool_calls` wire flag (issue #104).
+   */
+  supportsParallelTools?: boolean;
+  /**
    * Optional hot-swap supervisor. When provided, the loop re-probes
    * `/props` at the start of every turn and inspects the `modelId` of
    * each completion; if the operator swaps the model behind
@@ -577,6 +584,7 @@ export class AgentLoop {
             toolTransport: this.deps.toolTransport ?? "grammar",
             toolCallAdapter: this.deps.toolCallAdapter ?? null,
             supportsSlotAffinity: this.deps.supportsSlotAffinity ?? true,
+            supportsParallelTools: this.deps.supportsParallelTools ?? true,
             llmComplete: this.deps.llmComplete,
             ...(this.deps.llmCompleteStream
               ? { llmCompleteStream: this.deps.llmCompleteStream }
--- a/src/agent/step-executor.test.ts
+++ b/src/agent/step-executor.test.ts
@@ -13,6 +13,7 @@ import { buildGrammar } from "../llm/grammar/build-grammar.js";
 import { createEmptySessionState } from "../session/session-state.js";
 import { DEFAULT_TOOL_DESCRIPTORS } from "../prompt/tool-descriptors.js";
 import { replyTool } from "../tools/conversation/reply.js";
+import { resetConfigCache } from "../config/index.js";
 import type {
   CapabilitiesSummary,
   SkillCatalogEntry,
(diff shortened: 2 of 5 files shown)


plombeer31:
Reviewed while integrating this into the 0.3.3 build — it's in. One finding: as wired, the provider-capability half of this never fires for the provider the change is about.

**`src/llm/provider/registry/register-built-in-providers.ts:133` — `supportsParallelTools` is never `false` for Gemini.**

Every registry path sets it from `supportsTools`:

```ts
supportsParallelTools: entry.supportsTools ?? true,
```

…including the Gemini branch at :124-136. So for a Gemini provider that supports tools at all, the capability is `true`, and `buildLlmStreamParams` computes:

```ts
parallelToolCalls: getConfig().agent.maxParallelToolCalls > 1 && (true)
```

With the default `maxParallelToolCalls` of 8 that's `true`, so the wire still carries `parallel_tool_calls: true` and Gemini still streams parallel calls without stable indices — the symptom issue #104 describes. The only provider that declares `false` today is `subscription-cli-provider.ts:116`.

Net effect: the change is real but currently only bites when someone sets `maxParallelToolCalls=1`; the capability term is inert. The fix is presumably a per-entry flag (or a Gemini-specific `supportsParallelTools: false`) rather than deriving it from `supportsTools`.

The rest reads correctly, and the two `openai-build-body` cases pin the wire format nicely.

PR #212, opened 2026-08-21
fix(llm): default Gemini supportsParallelTools to false

### LINKED ISSUE #104: maxParallelToolCalls=1 still sends parallel_tool_calls: true to OpenAI-compatible providers

yablokolabs:
## Summary

Follow-up to PR #207 (fixes #104). The review on #207 flagged that the provider-capability half of the fix is currently inert for Gemini — every built-in provider kind sets `supportsParallelTools` from `supportsTools`, and since Gemini supports tools, the capability is always `true`. The only provider that declared `false` was `subscription-cli`.

This adds an explicit `supportsParallelTools?: boolean` field to the provider config entry and the user config schema. Every provider kind reads it with fallback to `supportsTools`, but **Gemini defaults to `false`** — overriding the fallback. Users can explicitly set `supportsParallelTools: true` on a Gemini entry to opt in if their deployment supports it.

## Changes

- `src/config/llm-config.ts` — new `supportsParallelTools?: boolean` field on `UserLlmProviderEntry` + parser
- `src/config/config-schema.ts` — schema acceptance for the new field
- `src/llm/provider/registry/provider-types.ts` — new field on `LlmProviderConfigEntry`
- `src/llm/provider/registry/register-built-in-providers.ts` — all kinds use `entry.supportsParallelTools ?? entry.supportsTools ?? true`; Gemini uses `entry.supportsParallelTools ?? false`

## Test coverage

7 new tests:
- Config: round-trip through `parseLlmProviderEntry` (explicit `true` and `false`), absent field stays `undefined`, non-boolean rejected
- Registry: Gemini defaults `supportsParallelTools` to `false`, explicit `true` override respected

## Verification

- `npm run lint` clean
- 909 tests pass across `src/llm/`, `src/agent/`, `src/config/`

### DIFF
--- a/src/config/config-schema.ts
+++ b/src/config/config-schema.ts
@@ -769,6 +769,7 @@ export interface AtomicAgentConfig {
       apiKeyHeader?: string;
       supportsTools?: boolean;
       supportsVision?: boolean;
+      supportsParallelTools?: boolean;
       requestTimeoutMs?: number;
       promptCache?: "auto" | "off" | "explicit-markers";
       providerPreferences?: Record<string, unknown>;
--- a/src/config/llm-config.test.ts
+++ b/src/config/llm-config.test.ts
@@ -505,4 +505,82 @@ describe("llm-config", () => {
       }),
     ).toThrow(/extraArgs\[1\]/);
   });
+
+  it("round-trips supportsParallelTools through parseLlmProviderEntry", () => {
+    const parsed = parseUserConfigFile({
+      version: USER_CONFIG_VERSION,
+      llm: {
+        activeTextProvider: "gemini",
+        activeEmbeddingProvider: "gemini",
+        toolTransport: "auto",
+        providers: [
+          {
+            id: "gemini",
+            kind: "gemini",
+            baseUrl: "https://generativelanguage.googleapis.com",
+            apiKey: "test-key",
+            defaultChatModel: "gemini-2.5-flash",
+            supportsParallelTools: false,
+          },
+          {
+            id: "openai",
+            kind: "openai-compatible",
+            baseUrl: "https://api.openai.com",
+            apiKey: "test-key",
+            defaultChatModel: "gpt-4o",
+            supportsParallelTools: true,
+          },
+        ],
+      },
+    });
+    const gemini = parsed.llm?.providers.find((p) => p.id === "gemini");
+    expect(gemini?.supportsParallelTools).toBe(false);
+    const openai = parsed.llm?.providers.find((p) => p.id === "openai");
+    expect(openai?.supportsParallelTools).toBe(true);
+  });
+
+  it("accepts undefined supportsParallelTools (absent field)", () => {
+    const parsed = parseUserConfigFile({
+      version: USER_CONFIG_VERSION,
+      llm: {
+        activeTextProvider: "gemini",
+        activeEmbeddingProvider: "gemini",
+        toolTransport: "auto",
+        providers: [
+          {
+            id: "gemini",
+            kind: "gemini",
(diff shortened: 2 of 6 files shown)


sosidudku1:
Thank you @yablokolabs, and apologies for the month without a reply. The change is clean and well tested, but main has moved underneath it.

The case for defaulting Gemini to `supportsParallelTools: false` was #104: Gemini's OpenAI-compatible stream sent parallel calls without an `index`, and our consumer merged them into one call. That root cause was fixed in ce68f353 for #103. `src/llm/provider/openai/openai-stream-consumer.ts` now keys index-less deltas by their `id`, so distinct calls stay distinct, and `GeminiProvider` inherits that path. With the fix in place, turning parallel calls off for every Gemini user would cost latency without fixing a bug we know of.

So we are closing this one. If you still see merged or dropped calls with Gemini on v0.6.4, please open an issue with a captured stream and we will pick it up. A standalone `supportsParallelTools` override per provider entry, with no Gemini default, would be a fine small PR if you hit a provider that needs it.

PR #214, opened 2026-08-21
fix(session): prune turns after each runTurn to prevent heap OOM

### LINKED ISSUE #121: TUI crashes with JavaScript heap OOM during long agentic sessions (25+ tool-call steps)
**SUMMARY:**
Environment: Windows 11 Home, PowerShell 7, atomic-agent v0.1.72, cloud mode (DeepSeek via OpenRouter)

Trigger: Single turn with 25 agentic steps reading large source files

Error: _FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory_ at ~4GB V8 heap
NDJSON trace, text export, and screenshot attached

Root cause: TUI retains full conversation + tool outputs in React/Ink state without conversation windowing or pruning

**FULL:**

The atomic-agent crashed with a
JavaScript heap out of memory (OOM) error after approximately 25+
tool-call steps and 40+ minutes of operation. The crash was caused by
unbounded growth of the conversation state in the agent's TUI, which
retained full conversation history, tool outputs, and internal state
without pruning.

Root Cause:
The atomic-agent's TUI (built with React/Ink) does not implement
conversation windowing or pruning. As the session progresses, the
agent accumulates:
- All user messages
- All agent responses (including internal reasoning)
- All tool outputs (including large JSON responses from MCP servers)
- Internal state (token counts, timing, model metadata)

This results in monotonic memory growth, eventually exhausting the
default 4GB Node.js V8 heap limit.

Evidence from Sanitized Trace:
The sanitized NDJSON trace (trace-sanitized.ndjson) shows:
- 215 total events spanning ~2400 seconds
- Monotonic increase in token counts across events (prompt_tokens,
completion_tokens)
- Tool outputs growing in size (e.g., web_search results, web_fetch
content, file reads)
- Final events show token counts in the tens of thousands before crash
- The crash event (FATAL ERROR) occurs after event #215

Specifically, the trace reveals that the agent's internal state
includes full copies of:
- Web search results (snippets, URLs)
- Web fetched content (full markdown/text)
- File contents (when using read_file or similar)
- Tool invocation details (parameters, results)

Impact:
- Loss of the current agent session, requiring restart
- Potential loss of unsaved work if not persisted elsewhere
- Need to re-establish context after restart
- Disruption to any ongoing automated tasks

Attachments:
1. trace-sanitized.ndjson - Sanitized NDJSON trace (215 events, all PII removed)
2. powershell-crash-sanitized.txt - Sanitized PowerShell crash output
(GC logs, FATAL ERROR, stack trace)
3. screenshot-sanitized.png - Sanitized screenshot of the agent TUI at
crash time

[powershell-crash-sanitized.txt](https://github.com/user-attachments/files/31000492/powershell-crash-sanitized.txt)
Image
[trace-sanitized.zip](https://github.com/user-attachments/files/31000501/trace-sanitized.zip)

yablokolabs:
## Summary

Fixes #121. `SessionState.turns[]` grows unbounded — every `assistant_tool_call`, `tool_result`, and `assistant_reply` is appended forever. Large tool outputs (file reads, web fetches) in a 25+ step session push the V8 heap past 4 GB.

The TUI side (`feed`, `messages`, `reasoning`) was already bounded via `pushRing` with `ringBufferSize`. The real culprit was `turns[]`.

## Fix

`packConversation` already computes the token-budgeted visible window at prompt-build time. This PR applies the **same windowing at the persistence boundary**: after `runTurn` returns, `pruneSessionTurns()` drops old turns beyond `conversationMaxTokens` and replaces them with a compact summary line. The next prompt build sees the pruned state; `packConversation` computes a fresh summary from what remains.

The pruning is zero-cost when turns already fit within budget (returns a copy of the input array). It only activates when the accumulated turns exceed the token cap.

## Changes

- `src/session/conversation-turn.ts` — new `pruneSessionTurns(turns, maxTokens)` pure function
- `src/runtime/bootstrap.ts` — call `pruneSessionTurns` before `sessionStore.save` in `executeTurn`
- `src/session/index.ts` — export the new function

## Test coverage

5 new tests:
- Within-budget passthrough (no pruning)
- Over-budget pruning with summary prepended
- Empty input
- Immutability guarantee (input array not mutated)
- Large-turn-list pruning

## Verification

- `npm run lint` clean
- 5043/5049 tests pass (2 pre-existing flaky: `send-message-concurrency`, `llm-health-poller`)

### DIFF
--- a/src/runtime/bootstrap.ts
+++ b/src/runtime/bootstrap.ts
@@ -157,6 +157,7 @@ import type { AgentLoopEvent, RunTurnResult } from "../agent/agent-loop.js";
 import {
   SessionStore,
   createEmptySessionState,
+  pruneSessionTurns,
   type SessionState,
 } from "../session/index.js";
 
@@ -2101,8 +2102,22 @@ export async function createAgentRuntime(
         maxSteps: runOptions.maxSteps ?? config.agent.maxSteps,
         signal: runOptions.signal ?? new AbortController().signal,
       });
-      sessionStore.save(result.session);
-      return result;
+      // Prune old turns to bound memory growth (issue #121). Without
+      // this, every tool_call / tool_result / assistant_reply is kept
+      // forever — large outputs (file reads, web fetches) push a long
+      // session past the V8 heap limit. packConversation already
+      // determines the visible window at prompt-build time; we apply
+      // the same windowing here so the persistent state stays bounded.
+      const prunedTurns = pruneSessionTurns(
+        result.session.turns,
+        config.agent.conversationMaxTokens,
+      );
+      const prunedSession =
+        prunedTurns.length === result.session.turns.length
+          ? result.session
+          : { ...result.session, turns: prunedTurns };
+      sessionStore.save(prunedSession);
+      return { ...result, session: prunedSession };
     });
   };
 
--- a/src/session/conversation-turn.test.ts
+++ b/src/session/conversation-turn.test.ts
@@ -8,6 +8,7 @@ import {
   renderTurnForPrompt,
   toolResultTurn,
   trimTurnsToTokens,
+  pruneSessionTurns,
   userTurn,
   type ConversationTurn,
 } from "./conversation-turn.js";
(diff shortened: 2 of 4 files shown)


sosidudku1:
Closing this in favour of #255, with the full reasoning in #121.

Short version: the crash evidence points away from conversation state. The process died at 4083 MB against Node's default ~4288 MB ceiling, a full mark-compact freed only ~11 MB, and the session was 6 turns / 30 steps with ~22 of its 34 minutes spent idle — growth tracked wall-clock rather than work. Measured directly, TUI reducer state retained ~0 MB and a 312-turn transcript under 1 MB. The likely retainer is the one #232 found: the shipped binary ran React's development build, which emits a `performance.measure()` per render that Node keeps for the life of the process, while the TUI redraws on a timer even when idle.

That does not make your patch wrong. Unbounded `SessionState.turns[]` is real, and #255 carries the same `conversation-turn.ts` windowing forward along with three other bounded paths and a heap-pressure warning. Both PRs touch that file, so they cannot both land — this is a merge-order decision, not a rejection of the work.

Thank you for the patch and, more importantly, for the NDJSON trace and crash log on #121. Those are what made the real diagnosis possible.

PR #284, opened 2026-08-29
feat: add /max_steps slash command to adjust max_steps at runtime

yablokolabs:
## Summary

Adds the TUI `/max_steps` slash command so operators can inspect or change the agent's per-turn step budget without restarting.

- `/max_steps` reports the active runtime value.
- `/max_steps <positive integer>` updates the live TUI orchestrator and persists `agent.maxSteps` to the user config.
- Persistence failures keep the runtime update and report a warning.
- Invalid values are rejected through the existing positive-integer config validation.
- The command is available from the slash-command menu and remains compatible with the upstream TUI selection-drag callbacks.

## Validation

- 60 focused TUI/config tests pass.
- `npm run lint` passes.
- `npm run build` passes.
- The full suite has one pre-existing/unrelated sidecar concurrency failure in `src/sidecar/send-message-concurrency.test.ts`.


sosidudku1:
Sorry for the long silence on this one, and thanks for the careful work: the runtime-only fallback when persisting fails and the validation reuse were both good touches.

main has since changed how step budgets work, and that overtakes this PR. Since v0.5.6, agent.maxSteps is the length of one leg, and a task keeps going with auto-continue up to agent.task.maxSteps (default 1000). In e4e519c7 (v0.6.3) the TUI stopped passing a per-turn maxSteps unless --max-steps is given (chat-orchestrator.ts:1181-1183). tui-command.ts:193-196 explains why: always passing the config value turned the leg into a hard ceiling and stopped every TUI turn at 25 steps. Because this PR passes maxSteps on every turn, rebasing it as is would bring that back.

So we're going to close this. If a live knob is still useful to you, we'd be glad to review a smaller version that edits agent.task.maxSteps (or the leg length) through config, without passing a per-turn maxSteps. Thanks again.

PR #290, opened 2026-08-31
feat(tui): steer an Ollama-shaped External URL into the provider wizard

plombeer31:
## What / why

Users keep pointing the **External llama.cpp** URL at Ollama (`:11434`). The `/health` probe classifies it `openai-compat` and refuses the save — since 26102be the verdict is at least visible on the panel's status line, but there was no path forward: acting on it meant retyping the same URL into the provider wizard four screens away.

Re-verified against a live `ollama serve` before coding: `/health` answers **404** and `/v1/models` answers the OpenAI list shape, so `checkLlamaServer`'s verdict for Ollama is exactly `openai-compat` (status 404, via the secondary `/v1/models` probe).

### The steer

- On an `openai-compat` refusal the External pane now opens a prompt naming the server ("Ollama detected — add it as a cloud provider?"). `y`/Enter deep-links into the provider wizard on the OpenAI-compatible route with the probed URL prefilled; `n`/Esc dismisses. The verdict line still lands on the feed + status line as before.
- A URL on Ollama's default port lands on the **"Ollama (local)" preset** — the exact state Enter on that `pick_kind` row builds (entry id, `OLLAMA_API_KEY` env var, no key screen since `ollama serve` is keyless, straight to the live `/v1/models` pick list) — but keeping the operator's own host, so a remote Ollama on `192.168.x.x:11434` survives.
- Any other compat server (LM Studio, vLLM, KoboldCpp) lands on the manual `openai-compatible` row's URL screen with the URL prefilled; Enter confirms and walks the normal URL → key → model flow (loopback keeps the key optional).
- `describeLlamaHealthFailure` now names Ollama outright for `:11434` URLs and points at the preset row, so the onboarding custom-URL branch (which shares the describer) stops calling an Ollama user's server merely "OpenAI-compatible".

Wizard Esc behavior comes for free: stepping back from the steered screen lands on the provider list with the cursor on the row the steer chose.

## Test evidence

New/extended vitest coverage, all of which **fails without the change** (verified by stashing the implementation: 7 failures, 0 with it):

- `src/tui/providers/openai-compat-steer.test.ts` — verdict-to-steer mapping: Ollama URL → `presetId: "ollama"`, `phase: "chat_model_line"`, host kept; non-Ollama → manual row at `base_url`, URL prefilled.
- `src/llm/describe-llama-health-failure.test.ts` — the Ollama-named line for `:11434`, `looksLikeOllamaUrl` accept/reject cases; the existing non-Ollama steer line still asserted.
- `src/tui/llm-panel/llm-panel-external.test.ts` — reducer open/close round-trip; `y` dispatches close + Cloud-pane flip + `providers_wizard_opened` with the prefilled wizard; `n`/Esc dismiss; panel hotkeys swallowed while the prompt owns the keyboard.

`npm run lint` clean; targeted suites green: `src/tui/llm-panel`, `src/tui/providers`, `llm-panel.test.tsx`, `providers-wizard.test.tsx`, `llm-mode-rows-cloud.test.tsx`, `llama-server-health.test.ts`, `slash-command-handler.test.ts`, `app-key-bindings.test.ts` — 33 files, 486 tests, 0 failures.

Reported on Discord: https://discord.com/channels/1515649306781155428/1515650562430079048/1542269245050589287

### DIFF
--- a/src/llm/describe-llama-health-failure.test.ts
+++ b/src/llm/describe-llama-health-failure.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, it } from "vitest";
-import { describeLlamaHealthFailure } from "./describe-llama-health-failure.js";
+import {
+  describeLlamaHealthFailure,
+  looksLikeOllamaUrl,
+} from "./describe-llama-health-failure.js";
 import type { HealthResult } from "./llama-server-health.js";
 
 function result(partial: Partial<HealthResult>): HealthResult {
@@ -21,6 +24,18 @@ describe("describeLlamaHealthFailure", () => {
     );
     expect(line).toContain("OpenAI-compatible");
     expect(line).toContain("openai-compatible, base URL http://127.0.0.1:1234");
+  });
+
+  it("names Ollama outright when the openai-compat server sits on :11434", () => {
+    // The most common shape of this verdict by far: the External URL
+    // pointed at `ollama serve`. "openai-compatible" alone did not tell
+    // an Ollama user the message was about them.
+    const line = describeLlamaHealthFailure(
+      result({ kind: "openai-compat", status: 404, error: "http 404" }),
+      "http://127.0.0.1:11434",
+    );
+    expect(line).toContain("Ollama");
+    expect(line).toContain("base URL http://127.0.0.1:11434");
   });
 
   it("says wait, not reconfigure, while the model is loading", () => {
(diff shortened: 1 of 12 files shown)


plombeer31:
Follow-up 47defe6 addresses the review findings:

- **Modal precedence**: the steer's key branch now sits after the External URL draft's, matching `LlmPanelModals` render order, and the reducer refuses to open the steer while the draft is open (the probe answers asynchronously; the editor the operator reopened meanwhile keeps the keyboard, and the next save re-probes and re-offers the steer). Regression tests cover both: the reducer skip, and — state built directly — that a visible editor receives `8`/`y` as text instead of the hidden steer swallowing them.
- **Vacuous test**: 'swallows panel hotkeys' now proves its baseline first (`]` cycles the pane without the prompt, dispatches nothing with it); verified it fails with the implementation reverted to the parent commit.
- **Remote Ollama text**: the :11434 refusal now routes non-loopback hosts through the manual openai-compatible row (which asks for the URL, so the host survives) and notes any API key value passes; the 'Ollama (local)' preset instruction — whose row never shows a base-URL screen — is reserved for loopback URLs, the only case where following it by hand saves the right server.

tsc clean; 92 test files / 963 tests green across llm-panel, providers, components, hooks, app-key-bindings, tui-command on Node 25.7.

PR #293, opened 2026-08-31
feat(local-llm): CPU llama.cpp backend fallback for Windows boxes whose GPU build cannot serve

plombeer31:
## What

Windows machines without a working GPU compute stack — typically iGPU-only boxes (AMD 5600G / Vega) — could not load local models at all: `windows-backend-variant.ts` only knew the Vulkan and two CUDA turboquant zips, so a box with no NVIDIA driver always got Vulkan, and when that build crashed or hung on model load there was **no CPU build to reach and no way to ask for one**. A `device: "cpu"` override does not help — that only sets `-ngl 0`, the broken compute backend is still baked into the binary.

This PR adds the full escape hatch:

- **CPU backend variant** — registers `llama-turboquant-windows-x64-cpu.zip`, which the `atomic-llama-cpp-turboquant-nightly` release already publishes (verified via the GitHub API against `turboquant-07b9908`, the only release; its asset list is macOS-arm64, linux-x64-vulkan, and win-x64 {vulkan, cuda-12.4, cuda-13.3, **cpu**}). Windows is the only platform with a CPU asset, so the fallback is Windows-only; if the nightly ever publishes a linux-x64-cpu zip the same machinery extends to it.
- **Selection logic unchanged for GPUs** — detection still prefers CUDA > Vulkan; the CPU build is never chosen by detection.
- **Automatic fallback on start failure** — when a GPU-variant install spawns but never becomes healthy (new typed `DaemonHealthError`, i.e. the server crashed or could not load the model — pre-spawn failures like a missing model file do not trigger it), both the TUI start path and CLI `models start`:
  1. surface a clear message,
  2. persist `localModels.managed.backendVariant: "cpu"` (config v46) — without this the next auto-update's variant-staleness check would resolve Vulkan again and reinstall the build that just failed,
  3. swap in the CPU zip (stopping the half-started daemon first, or the Windows file lock wedges the swap), and
  4. retry the start once (never recursing further).
- **User-selectable override** — `localModels.managed.backendVariant`: `"auto"` (default, previous behavior) | `"cpu"` | `"vulkan"` | `"cuda-12.4"` | `"cuda-13.3"`. Pinning a GPU variant is also how you undo the automatic fallback after a driver fix. Wired into the config-free local-llm layer via the same push-in pattern as `setCustomLocalModels`.

Why no "no capable GPU detected → CPU at selection time": there is no reliable pre-download Vulkan-capability probe on Windows (nvidia-smi covers NVIDIA only, and the reporter's box *has* a Vulkan device — it just cannot load a model on it), so eligibility is failure-driven plus the explicit override.

**On the linked upstream issue**: ggml-org/llama.cpp#22786 is a Gemma-4 tool-call-parsed-as-content bug, closed 2026-05-08 — unrelated to Vulkan model-load failures, and the nightly build (`turboquant-07b9908`, published 2026-08-14) postdates it anyway. There is also no pinned tag to bump in this repo: the installer always resolves the newest nightly release, so no version bump ships here. If the reporter's crash is an upstream ggml-vulkan bug, the durable fix is a fresh nightly build; this PR makes such boxes work either way.

## Test evidence

- `npm run lint` (tsc) clean.
- New/extended vitest coverage, each proven to fail without the change (verified by temporarily reverting the specific behavior):
  - `windows-backend-variant.test.ts` — CPU zip registration, override pins without probing nvidia-smi, override set *after* a cached detection still wins, `"auto"` restores detection, non-Windows platforms ignore the preference, `isWindowsGpuBackendAsset` matrix (4 failed with the override check disabled).
  - `daemon-lifecycle.test.ts` — `startDaemon` rejects with typed `DaemonHealthError` when the spawned server never becomes healthy (failed when reverted to a bare `Error`).
  - `cpu-backend-fallback.test.ts` — full eligibility matrix (platform / pinned variant / already-CPU / non-health errors) and the swap glue (stop-before-download ordering, stop failure tolerated, download failure propagated, progress/signal forwarded).
  - `backend-installer.test.ts` — configured `"cpu"` over an installed Vulkan build reports `updateAvailable` (the anti-flip-back loop guard).
  - `config-schema.test.ts` — v46 `backendVariant` default (incl. transparent v45 upgrade), explicit value preserved, unknown value rejected.
- Full `src/local-llm`, `src/config`, `src/tui/local-models`, `src/cli/models-command` suites pass (the pre-existing unhandled `spawn EACCES` warning in `local-models-orchestrator-auto-update.test.ts` reproduces on clean `main`).

## Windows QA needed

Runtime verification requires an iGPU-only Windows box (none available here). Manual steps are in the PR discussion below the fold: run `atomic-agent models start` on a box where Vulkan fails to load a model, expect the "falling back to the CPU build" flow, `backendVariant: "cpu"` written to config.json, and a healthy daemon on the CPU build; subsequent starts must stay on the CPU zip.

Reported on Discord: https://discord.com/channels/1515649306781155428/1515649308161085612/1540671929755631699 (+ follow-ups)

## Review fixes (5d0a242)

- **TUI fallback download now has a deadline and progress**: `fallBackToCpuBackendAndRetry` was calling `fallBackToCpuBackend(dataDir)` bare — a stalled-open connection could pin the start on "starting" forever with zero feedback. It now passes `AbortSignal.timeout(BACKEND_DOWNLOAD_TIMEOUT_MS)` and surfaces the download as a regular backend pull (`started`/`progress`/`finished`/`failed`) on the bus, matching the auto-update path and the CLI path.
- **Fallback wiring is now integration-tested** (previously only the extracted pure pieces were): `local-models-orchestrator-cpu-fallback.test.ts` (5 tests — swap+persist+single retry, recursion cap, failed download reported, non-health and already-CPU cases inert, deadline+progress forwarded) and `models-handlers.test.ts` (3 tests — CLI retry lands on `--device cpu`, variant persisted, signal/progress forwarded, pre-spawn failures and failed downloads exit non-zero). Both proven non-vacuous: deleting either caller's retry block fails 5 tests; making the TUI download call bare again fails 1.
- Manual QA comment corrected: `models start` takes no model argument — step 1 now selects the model via `models pull` + `models use`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,8 @@ Managed mode downloads the backend, pulls GGUF models, selects the active model,
 
 The managed chat daemon stops when the last session exits, freeing the RAM and VRAM the model was holding; set `localModels.managed.stopOnExit: false` in `config.json` to keep the model warm between sessions. Daemons started standalone with `models start` are never touched.
 
+On Windows the backend zip is picked per machine (CUDA when a capable NVIDIA driver is present, Vulkan otherwise). If the GPU build cannot serve a model on your hardware — typical for iGPU-only boxes — the start falls back to the CPU build automatically and records `localModels.managed.backendVariant: "cpu"` in `config.json`; set it to `"auto"`, `"vulkan"`, `"cuda-12.4"` or `"cuda-13.3"` to pick a build yourself (e.g. after a driver update).
+
 Cloud models are searchable from the same command — by id, vendor, or capability, across every configured cloud provider:
 
 ```bash
--- a/src/cli/models-handlers.test.ts
+++ b/src/cli/models-handlers.test.ts
@@ -1,0 +1,167 @@
+import { mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../local-llm/index.js", async () => {
+  const actual =
+    await vi.importActual<typeof import("../local-llm/index.js")>(
+      "../local-llm/index.js",
+    );
+  return {
+    ...actual,
+    maybeAutoUpdateBackend: vi.fn(),
+    resolveManagedDevice: vi.fn(),
+    startChatAndEmbeddingDaemons: vi.fn(),
+    fallBackToCpuBackend: vi.fn(),
+  };
+});
+
+import { getUserConfigPath, writeUserConfigFileSync } from "../config/config-file.js";
+import { USER_CONFIG_DEFAULTS } from "../config/config-schema.js";
+import { getConfig, resetConfigCache } from "../config/index.js";
+import * as localLlm from "../local-llm/index.js";
+import { resolveBackendDir } from "../local-llm/index.js";
+import { writeBackendVersion } from "../local-llm/backend-version.js";
+import { DaemonHealthError } from "../local-llm/daemon-lifecycle.js";
+import {
+  WINDOWS_BACKEND_ASSETS,
+  setConfiguredBackendVariant,
+} from "../local-llm/windows-backend-variant.js";
+import { runLocalModelsStart } from "./models-handlers.js";
+
+const healthError = () =>
+  new DaemonHealthError(
+    "llama-server did not become healthy within 30000ms. Log tail:\n(no log)",
+  );
+
+/**
+ * Integration tests for the CPU-backend fallback retry block inside
+ * `runLocalModelsStart` — the CLI twin of the orchestrator wiring
+ * covered in `local-models-orchestrator-cpu-fallback.test.ts`. The pure
+ * pieces have their own tests; these prove the handler actually
+ * consults them, retries on the CPU device, and persists the variant.
+ */
+describe("runLocalModelsStart CPU-backend fallback", () => {
+  let stateDir: string;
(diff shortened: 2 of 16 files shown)


plombeer31:
**Manual QA steps (iGPU-only Windows box, e.g. AMD 5600G, no NVIDIA driver):**

1. Install this build, remove any previous backend install (delete the `backend/` dir under the local-models data dir), ensure `localModels.managed.backendVariant` is absent or `"auto"` in config.json, and select a model: `atomic-agent models pull qwen-3.5-4b && atomic-agent models use qwen-3.5-4b`.
2. Run `atomic-agent models start` (the model id comes from config; or start the model from the TUI's Local Models panel). Expect: Vulkan zip installs first (unchanged detection), the server fails to become healthy, and a clear message that it is falling back to the CPU build — not a raw crash. The CPU zip download must show live progress (CLI progress bar / TUI backend pull row), not a silent hang.
3. Expect the CPU zip (`llama-turboquant-windows-x64-cpu.zip`) to download, the daemon to become healthy, and the model to answer a prompt.
4. Open config.json: `localModels.managed.backendVariant` must now be `"cpu"`.
5. Restart the app and start the model again: it must go straight to the CPU build — no Vulkan reinstall from the auto-updater.
6. Optional override check: set `backendVariant: "vulkan"` manually and restart — the Vulkan build must be reinstalled (this is the documented way to undo the fallback after a driver fix).

PR #308, opened 2026-09-01
fix(cli): default NODE_ENV to production for unbundled runs

### LINKED ISSUE #307: Long-running TUI process dies with V8 heap OOM (~11 h): development build of `react-reconciler` floods the global performance buffer when running from `dist` (`NODE_ENV` unset)
## Summary

When the TUI is run the way the README documents for source deployments — `npm install` → build → `node dist/cli/index.js tui` — `process.env.NODE_ENV` is undefined. `react-reconciler`'s CommonJS entry then resolves to its **development** build, which instruments every React commit with `performance.mark`/`performance.measure`. A TUI re-renders continuously (timers, polling indicators, spinners), and nothing in the runtime ever drains the global performance buffer. The measure entries accumulate without bound until the V8 heap (~4 GB default) is exhausted and Node aborts the process:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

In production this presents as: an unattended agent works fine all morning, then goes completely silent around the 11-hour mark. Process, pane, and (if it is the only session) the tmux server are gone; there is no stack in the application log because the abort is native.

## Environment

- atomic-agent **v0.4.2** (local build; dependency tree and rendering-related code verified byte-identical to the vanilla tag — see *Local-build audit* below)
- Node **v24.19.0**, Linux x64
- Deployment: clone → `npm install` → build → `node dist/cli/index.js tui` inside tmux, one channel active (Telegram), default state dir (`~/.atomic-agent`)
- No `NODE_ENV` set anywhere in the launch environment (this is the default outcome of following the README)

## What happens (production evidence)

| Uptime | Event |
|---|---|
| ~7–8 h | `(node:<pid>) MaxPerformanceEntryBufferExceededWarning: Possible perf_hooks memory leak detected. 1000001 measure entries added to the global performance entry buffer. Use performance.clearMeasures to clear the buffer.` |
| ~10 h 58 m | Repeated GC failure logs, then: |

```
<--- Last few GCs --->

[<pid>:0x...] 39492074 ms: Scavenge (interleaved) 4090.5 (4099.2) -> 4086.9 (4099.2) MB, pooled: 0 MB, 6.40 / 0.03 ms  (average mu = 0.327, current mu = 0.312) allocation failure;
[<pid>:0x...] 39493195 ms: Mark-Compact (reduce) 4091.7 (4100.2) -> 4086.4 (4092.5) MB, pooled: 0 MB, 634.51 / 0.28 ms  (+ 376.4 ms in 76 steps since start of marking, biggest step 5.4 ms, walltime since start of marking 1021 ms) (average mu = 0.32

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 0x... node::OOMErrorHandler(char const*, v8::OOMDetails const&) [node]
 2: 0x...  [node]
 ...
```

- Heap pinned at ~4.09 GB (the V8 default ceiling for 64-bit Node), runtime 39,492 s ≈ **10 h 58 m**.
- Corroborated by the OS: `systemd-coredump: Process <pid> (MainThread) of user <uid> dumped core.` seconds after the abort.
- The growth rate implied by the warning (1,000,001 entries by ~7–8 h) is roughly 35–40 measure entries per second — consistent with a continuously redrawing Ink TUI times the dev reconciler's instrumentation call sites.

## Root cause

**1. The dev/prod switch depends on `NODE_ENV`, which is unset in this deployment mode.**

`node_modules/react-reconciler/index.js`:

```js
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./cjs/react-reconciler.production.js');
} else {
  module.exports = require('./cjs/react-reconciler.development.js');
}
```

The repository is aware of this hazard — `scripts/bundle-sea.ts` exists precisely to define it away for the SEA binary:

```ts
// `process.env.NODE_ENV === "production" ? prod : dev` switch. A
// NODE_ENV — which is every machine.
…
"process.env.NODE_ENV": JSON.stringify("production"),
```

…but the README-documented source/dist path (`npm install` … `node dist/cli/index.js tui`) leaves `NODE_ENV` undefined, so **every non-SEA deployment runs the development reconciler**.

**2. The development reconciler instruments every commit.**

- `react-reconciler.development.js`: **8** `performance.measure(` call sites
- `react-reconciler.production.js`: **0**

A TUI is a worst case for this: Ink re-renders on every timer tick, polling status change, and spinner frame — indefinitely.
**3. Nothing ever drains the buffer.**

There is no `performance.clearMeasures()`, `performance.clearResourceTimings()`, or `PerformanceObserver`-based drain anywhere in the runtime source, so the global performance entry buffer grows for the life of the process. The entries are referenced objects; GC cannot reclaim them. At ~1 M entries the heap is exhausted (the Node warning fires at the 1,000,001st entry; the fatal OOM follows once the accumulated entries plus live heap reach the ceiling).

## Reproduction

**Tier 1 — deterministic, seconds (verified locally):**

```sh
node --max-old-space-size=64 dist/cli/index.js tui
```
Crashes with the identical `FATAL ERROR: Ineffective mark-compacts near heap limit` during initial rendering — within ~2 s of launch, with the same native `OOMErrorHandler` stack as the production crash. (At 64 MB the boot-time measure flood alone exhausts the ceiling; no channel or traffic needed.)

**Tier 2 — minutes (recommended for maintainer verification):**

```sh
node --max-old-space-size=256 dist/cli/index.js tui
```

Boots, then OOMs the same way once steady-state rendering accumulates enough entries. Exact time depends on redraw frequency.

**Tier 3 — real-world defaults (production evidence above):**

Default heap (~4 GB), one active channel: silent death at roughly the 11-hour mark, preceded by the `MaxPerformanceEntryBufferExceededWarning` a few hours in.

**Control (confirms attribution):** same launch with `NODE_ENV=production` selects `react-reconciler.production.js` (zero measure call sites); no warning, no growth, no crash.

## Suggested fixes

1. **Default the dist/CLI entry to production React**: set `NODE_ENV=production` in the CLI entry (before anything requires `ink`/`react-reconciler`) when it is unset — mirroring what `scripts/bundle-sea.ts` already does for the SEA path. This is the actual root-cause fix for the renderer instrumentation.
2. **Defense-in-depth drain**: a periodic `performance.clearMeasures()` (e.g., every few minutes, or when `performance.getEntriesByType('measure').length` exceeds a threshold) — cheap, safe (the entries are diagnostics nobody consumes), and covers any other emitter.
3. Optionally, document that source/dist deployments should set `NODE_ENV=production`.

## Workaround (for anyone hitting this today)

```sh
NODE_ENV=production node dist/cli/index.js tui
```

## Impact

Any long-running source/dist deployment — regardless of provider, channel, or model — silently dies roughly 11 hours after boot. Until then it is fully healthy, which makes the failure easy to misattribute to the environment (in my case it initially masqueraded as an unrelated "agent stopped responding" incident). A supervised deployment masks the symptom but restarts every ~11 h; the leak itself is unconditional.

plombeer31:
## What

Adds `src/cli/env-defaults.ts` — a side-effect module that defaults `NODE_ENV` to `"production"` when unset — imported **first** in `src/cli/index.ts`, before the `tuiCommand` import chain loads `react-reconciler`. An explicit `NODE_ENV` (`development`, `test`) still wins. Also updates the `terminal-restore.ts` heap-OOM comment to name both defences.

## Why

Running the TUI from a source build (`node dist/cli/index.js tui`, `npm run cli`, `dev:cli`) leaves `NODE_ENV` unset, so `react-reconciler` resolves to its **development** build, whose Component Performance Track calls `performance.measure()` on every React commit. Node keeps user-timing entries alive for the life of the process and nothing drains the buffer, so a continuously redrawing TUI grows the heap without bound until V8 aborts: `FATAL ERROR: Ineffective mark-compacts near heap limit` at roughly the 11-hour mark — a silent native death (no JS runs after it) that takes the tmux server down with it when it is the only session.

`scripts/bundle-sea.ts` already defines `NODE_ENV` away for the SEA binary (same hazard, same words in its comment); the tsc-compiled `dist` path had no equivalent. The import has to be first because ESM evaluates imports depth-first before the importer's body — a plain assignment in the entry body runs after the reconciler has already picked its build.

## Test evidence

- `npm run lint` clean; `npx vitest run src/cli/env-defaults.test.ts src/cli/bin-alias.test.ts` — 7/7 pass. New tests cover both env branches plus a source-level guard that the import stays first (import order leaves no runtime trace).
- Installed `react-reconciler@0.33.0`: **8** `performance.measure(` call sites in `cjs/react-reconciler.development.js`, **0** in `cjs/react-reconciler.production.js`.
- Live TUI boot under a PTY at `--max-old-space-size=64` with a `--import` probe sampling the global performance buffer:
  - fixed run (`NODE_ENV` unset): reports `env=production`, **0 measure entries**, flat, survives.
  - control (`NODE_ENV=development`): measure entries grow monotonically (idle onboarding screen, ~2.5/s here; last sample name is React measuring Ink's `Box`), nothing ever drains them — the reporter's ~35–40/s with an active channel extrapolates to the observed ~11 h death at the 4 GB default ceiling.

Fixes #307

### DIFF
--- a/src/cli/env-defaults.test.ts
+++ b/src/cli/env-defaults.test.ts
@@ -37,6 +37,12 @@ describe("env-defaults", () => {
     expect(process.env.NODE_ENV).toBe("production");
   });
 
+  it("treats an empty NODE_ENV as unset", async () => {
+    process.env.NODE_ENV = "";
+    await import("./env-defaults.js");
+    expect(process.env.NODE_ENV).toBe("production");
+  });
+
   it("leaves an explicit NODE_ENV alone", async () => {
     process.env.NODE_ENV = "development";
     await import("./env-defaults.js");
--- a/src/cli/env-defaults.test.ts
+++ b/src/cli/env-defaults.test.ts
@@ -1,0 +1,54 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { readFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+/**
+ * `env-defaults.ts` keeps `dist` runs on the production React build
+ * (issue #307: the development reconciler floods the global performance
+ * buffer until V8 aborts). Two things have to stay true, and only the
+ * second is visible to a unit test:
+ *
+ *  1. an unset `NODE_ENV` becomes `"production"`, an explicit one wins;
+ *  2. the module is evaluated before `react-reconciler` — which means
+ *     it must be the first import of the CLI entrypoint, checked here
+ *     against the source because import order leaves no runtime trace.
+ */
+const repoRoot = resolve(dirname(fileURLToPath(import.meta.url)), "..", "..");
+
+describe("env-defaults", () => {
+  const originalNodeEnv = process.env.NODE_ENV;
+
+  beforeEach(() => {
+    vi.resetModules();
+  });
+
+  afterEach(() => {
+    if (originalNodeEnv === undefined) {
+      delete process.env.NODE_ENV;
+    } else {
+      process.env.NODE_ENV = originalNodeEnv;
+    }
+  });
+
+  it("defaults an unset NODE_ENV to production", async () => {
+    delete process.env.NODE_ENV;
+    await import("./env-defaults.js");
+    expect(process.env.NODE_ENV).toBe("production");
+  });
+
+  it("leaves an explicit NODE_ENV alone", async () => {
+    process.env.NODE_ENV = "development";
+    await import("./env-defaults.js");
+    expect(process.env.NODE_ENV).toBe("development");
+  });
+
+  it("is the first import of the CLI entrypoint", () => {
+    const source = readFileSync(
+      resolve(repoRoot, "src", "cli", "index.ts"),
+      "utf8",
+    );
+    const firstImport = source.match(/^import .*$/m)?.[0];
+    expect(firstImport).toBe('import "./env-defaults.js";');
+  });
+});
--- a/src/cli/env-defaults.ts
+++ b/src/cli/env-defaults.ts
@@ -24,6 +24,11 @@
  * the ordering.
  *
  * An explicitly exported `NODE_ENV` — `development` for React
- * debugging, `test` under a runner — is left alone.
+ * debugging, `test` under a runner — is left alone. An *empty* value
+ * (`NODE_ENV=` in a shell export or dotenv line) counts as unset: the
+ * reconciler switch compares `=== "production"`, so an empty string
+ * would silently pick the development build all the same.
  */
-process.env.NODE_ENV ??= "production";
+if (!process.env.NODE_ENV) {
+  process.env.NODE_ENV = "production";
+}
--- a/src/cli/env-defaults.ts
+++ b/src/cli/env-defaults.ts
@@ -1,0 +1,29 @@
+/**
+ * Defaults `NODE_ENV` to `"production"` before React can read it.
+ *
+ * React ships as two builds behind a runtime
+ * `process.env.NODE_ENV === "production" ? prod : dev` switch, and the
+ * development reconciler calls `performance.measure()` for every
+ * component render. Node keeps every user-timing entry alive for the
+ * life of the process, and the TUI redraws on a timer even while idle,
+ * so a `dist` run with `NODE_ENV` unset — the default outcome of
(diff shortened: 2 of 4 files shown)


plombeer31:
Closing as a duplicate of #310 — same fix from a concurrent session; #310 carries on.

PR #329, opened 2026-09-04
feat(integrations): move Telegram credentials into the hub

plombeer31:
> **Stacked on #328** (which is stacked on #327). Base is `feat/integrations-hub`, so the diff here is just Telegram. Retarget down the stack as each lands.

## What

Telegram becomes the hub's second tenant, so an operator setting up a new install has one place to paste every key rather than needing to know which service happens to own a tab.

## Scope: the credential only

Pairing, the owner id, start/stop and the live chat view **stay on the Telegram tab**. Those act on a running channel rather than configuring one, and folding a pairing countdown into a credential list would make both surfaces worse.

The hub owns *"what is my token"*; the tab owns *"what is the bot doing"*. The hub's summary line points at the tab, so the split is discoverable rather than something the operator has to work out.

## Validation at entry

The token is checked against the BotFather shape the scrubber already keys off — `<6..12 digits>:<≥30 [A-Za-z0-9_-] chars>`. A truncated paste stops being a silent "channel won't start" and becomes an immediate, specific complaint about the paste.

## Live status

`IntegrationStatusContext` gains `channelStates` alongside `mcpServerStates`, so three states that used to look identical now read differently:

| channel | badge | detail |
|---|---|---|
| (none / `disabled`) | `configured` | token saved — pair and enable on the Telegram tab |
| `up` | `connected` | channel up |
| `down` | `error` | channel failed to start — see the Telegram tab |

A merely *disabled* channel is deliberately **not** an error — that is a normal resting state, and badging it red would train operators to ignore the badge.

Telegram resolves its token at construction, so it declares `appliesLive: false` and the pane says a restart is needed rather than leaving the operator to guess why nothing happened.

## Testing

`npm run lint` and `npm test` green (7211 tests). 6 new tests; every existing Telegram test still passes untouched.

Verified end-to-end against a real runtime:

- `hunter2` → refused: *"Doesn't look like a bot token — expected digits, a colon, then a long string, as @BotFather issues it."* Nothing written.
- a well-formed token → saved, masked as `••••…+13`, badge flips to `configured` with the "pair and enable" detail
- it lands in `<stateDir>/.env` and is visible in `process.env` for the Telegram tab's own has-token check, without a restart

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1780,6 +1780,7 @@ Locked invariants (pinned by [src/integrations/integration-secrets.test.ts](src/
 4. **Edit mode swallows the whole keyboard.** `d`, `e` and `r` are bindings on this tab; inside the editor they are key material. A paste that silently triggered "clear field" halfway through would be both baffling and destructive.
 5. **A re-sync never yanks the cursor.** Rows are re-read on every refresh; the reducer clamps the selection instead of resetting it, so a background refresh cannot move the operator's place mid-edit.
 6. **Changing a Composio key drops the cached tool-router session.** A session belongs to the key that created it; reusing it across a key swap would keep talking to the old account. The orchestrator unmounts, clears the cache, re-resolves, and remounts live.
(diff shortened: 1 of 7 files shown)


plombeer31:

PR #331, opened 2026-09-07
feat(discord): Discord remote-control channel

plombeer31:
> **Stacked on #330 → #329 → #328 → #327.** Retarget down the stack as each lands.

## What

A Discord bot that relays DMs and @mentions to the agent and posts replies back — the same shape as the Telegram channel, sharing its `ChannelStatus` contract so the runtime, the TUI and the Integrations hub treat both alike.

## Why no client library

The channel talks to Discord over the raw HTTP + Gateway APIs using Node's built-in `fetch` and `WebSocket`. It needs six REST calls and one WebSocket state machine; `discord.js` would add a large transitive tree to a project that ships a single-file SEA binary, for code we would still have to wrap. Same reasoning as declining `@composio/core` in #327. **Net new dependencies: zero.**

## Safety-relevant design

- **Addressed messages only.** In a guild the bot acts only when @mentioned; in a DM, always. Its own posts and other bots are dropped outright — two agents in one guild would otherwise talk to each other forever.
- **`MESSAGE_CONTENT` is never requested.** Discord delivers full content for DMs and for messages that mention the bot without the privileged intent. So the operator never has to enable one (or pass verification at 100+ guilds), *and* the bot is structurally unable to read guild chatter it was not addressed in — the right default for something wired to a machine's shell.
- **One owner.** Unpaired means every message is refused. Pairing deliberately runs *before* the owner check, because claiming the first eligible message is what a pairing window is for.
- **`ownerUserId` is a string.** Snowflakes exceed `Number.MAX_SAFE_INTEGER`; parsing one as a number silently corrupts the last digits and would let the wrong account through.
- **Approvals prompt where the turn came from.** Without a bridge, `ApprovalRouter` falls through to the host handler — a TUI dialog on the operator's machine that someone driving from their phone cannot see, so the turn would block until it timed out. The bridge posts a two-button prompt in the originating channel, never approves on its own, and ignores presses from anyone but the owner.
- **One process per token.** Two gateways on one token receive every event twice and would run every turn twice, side effects included. Discord does not prevent this the way Telegram's 409 on parallel `getUpdates` does, so a lockfile has to.
- **Reconnect, but not forever.** Full-jitter backoff and RESUME where Discord allows it; 4004 (bad token) and 4014 (disallowed intents) stop the loop rather than burning the per-day session-start budget on a retry that can never succeed.

## Hub change: config-backed fields

The hub could previously only write `.env`, so Discord's owner id would have meant hand-editing `config.json` — half a setup surface. `IntegrationField` gains `store: "env" | "config"`; a config field carries `configPath` and no `envVar`, an env field the reverse, and the registry test pins that, because a field reading one store and writing the other fails silently.

## What is *not* here

No Discord TUI tab. Telegram has one for pairing and live chat state; Discord's equivalent is its own piece of work. Everything needed to set the channel up is in the Integrations hub, and the status line names the exact remaining command rather than pointing at a tab that doesn't exist:

```
Discord  · ready — enable with: atag config set '{"discord":{"enabled":true}}'
    Bot token=••••••••••••••••••••••••••••••••+30 | Owner user ID=123456789012345678
```

## Testing

`npm run lint` and `npm test` green (7308 tests). 67 new tests across six files.

- **REST client against real `discord.com`**: a syntactically valid but fake token returns `DiscordApiError 401` with *"Discord rejected the bot token (HTTP 401). Check it in the Integrations tab."*
- **Gateway state machine** driven with a fake socket through identify, resume-after-drop (asserting RESUME rather than a fresh IDENTIFY), non-resumable INVALID_SESSION, off-cycle heartbeat, malformed frame, fatal close, and clean shutdown.
- **Channel lifecycle**: disabled without a token (not `down`), lock released when the token is rejected, and `lastError` asserted never to contain the token.
- **Hub end-to-end against a real runtime**: a pasted client secret is refused (*"that's the shape of the client secret or public key. Use Bot → Reset Token."*), a malformed snowflake is refused, and valid values land in `.env` and `config.json` respectively.

One note on the first push: GitHub's push protection rejected it over a *fabricated* token fixture in a test — shaped realistically enough to trip the scanner. Fixed at the source by assembling the fixture at runtime rather than using the unblock link, so no token-shaped literal exists in the tree.


plombeer31:

PR #332, opened 2026-09-07
feat(integrations): move all Telegram controls into the hub, remove the tab

plombeer31:
> **Stacked on #331 → #330 → #329 → #328 → #327.** Retarget down the stack as each lands.

## What

Keeping a Telegram tab *and* a Telegram entry in the Integrations hub meant two places to configure one thing, and an operator having to know which owned what. **The tab is removed**; everything it did now lives in the hub.

```
Telegram  · token saved — press p to pair
  https://core.telegram.org/bots#botfather

> Bot token  ••••••••••••••••••••••••••••••••+13
    From @BotFather. Then press p to pair your account.
  Owner  —
  Channel  off

  p pair  ·  s restart
```

## Two additions to the hub contract

- **`kind: "boolean"`** — rendered as an on/off toggle, stored as a real boolean. This is how a channel's kill switch reaches the hub; without it an operator would set a token here and still need the CLI to turn the thing on.
- **`actions`** — descriptor-declared verbs beyond editing fields. Pairing an owner and restarting after a token change have to be doable in the same place, or the hub is half the story.

## No reimplementation

`IntegrationsOrchestrator` **delegates** to `TuiTelegramOrchestrator`, which already owns pairing windows, token writes and restarts. One implementation, so the two cannot drift. Only the tab's *UI* is deleted (`components/`, `telegram-key-bindings.ts`); the state, actions, reducer and orchestrator stay.

An action that cannot work isn't offered: `available()` hides `pair`/`restart` until a token exists, because pairing opens a window that claims the next DM — with no bot to DM it could only ever time out.

Discord gets the same treatment for consistency: a Channel toggle and a restart action, so its status line no longer has to tell the operator to go run a config command by hand.

## Nothing lost

`/telegram` and all its verbs still work and now land on Integrations, as does <kbd>ctrl+g</kbd> <kbd>g</kbd> — the menu node is kept as a deep link so anyone looking for "Telegram" still finds it. The only verb dropped is `/telegram token`, which opened the tab's modal; the hub's field replaces it.

## A real bug the tests caught

A row with no `actions` array crashed the key handler and would have taken the whole tab down. Guarded, with the fixtures that exposed it.

## Testing

`npm run lint` and `npm test` green (7310 tests).

Verified against a real runtime: toggling Channel writes `telegram.enabled` **and** calls `setEnabled` on the channel; `p` opens a pairing window; `s` restarts; the detail view shows token (masked), owner, channel state and both verbs. Confirmed in a real 120×40 TUI that the Telegram tab is gone from the Manage strip.


plombeer31:

PR #333, opened 2026-09-07
fix(telegram): report a dead polling loop instead of staying "up"

plombeer31:
> **Stacked on #332 → #331 → #330 → #329 → #328 → #327.** Retarget down the stack as each lands. The fix itself is independent of the stack and can be cherry-picked onto `main` if you'd rather land it first.

## The bug

`bot.start()` is fire-and-forget, and the grammy adapter threw away the promise it returns:

```ts
void bot.start({ onStart }).catch(() => undefined);
```

That promise settles when polling **ends** — resolving on a clean stop, rejecting on a fatal error. Discarding it meant the channel could never learn its poller had died. So after, say, Telegram's 409 (a second process polling the same token), the channel sat at `up`:

- every message afterwards went **silently unanswered**
- the TUI badge stayed **green**
- the rejection was **unhandled**

## How I found it

While testing the Telegram channel for real. One of my own diagnostic `getUpdates` calls took the poller down; the channel reported `up` for several minutes while messages queued at Telegram (`pending_update_count: 1` and climbing). Nothing in the product would have told an operator why their bot had gone quiet.

## The fix

`BotInstance.start` gains an `onStopped(error?)` callback. The adapter wires it to the settlement instead of swallowing it, and the channel:

- transitions to **`down`** with a scrubbed reason
- **releases the lock**, so a restart can re-acquire rather than being permanently unstartable in that process
- clears the cached bot identity

Two guards keep it honest:

- `stopRequested` is set before a deliberate `stop()`, so a normal shutdown still settles as `disabled` and never flashes `down`
- the callback is checked against the current bot instance, so a late callback from a previous generation (restart, token change) can't knock down the live one

## Testing

`npm run lint` and `npm test` green (7315 tests). Four new cases:

| case | asserts |
|---|---|
| poller dies with an error | `down`, reason contains `Conflict`, statuses `starting → up → down` |
| poller dies with no error | `down`, reason `polling stopped unexpectedly` |
| lock released on death | `release()` called, and a subsequent `start()` re-acquires and reaches `up` |
| deliberate `stop()` | settles `disabled`, never emits `down` |
| token in the death message | scrubbed to `<token>` |

**Verified non-vacuous**: with the `onStopped` wiring removed, 4 of the 5 fail. (This repo has a history of tests that pass against nothing, so I checked.)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1766,6 +1766,10 @@ Locked invariants (pinned by [src/tui/mcp/mcp-reducer.test.ts](src/tui/mcp/mcp-r
 4. **The editor is disabled on the MCP tab while a modal is open.** `mcpTabBusy` in `app-key-bindings.ts` covers both `addModal !== null` (lets the `MultiLineEditor` capture every keystroke) and `removeConfirm !== null` (claims the `y`/`n` confirmation keys against the global nav cycler).
 5. **Variant γ surface is opt-in but on by default.** Restarting the runtime is no longer required after add/remove — the prompt's `### tools` catalog and GBNF grammar are rebuilt on the next step. KV-cache for in-flight sessions is invalidated once per add/remove (the persona stays byte-stable; only the rendered tools block changes).
 
+### Poller liveness
+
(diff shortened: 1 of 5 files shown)


plombeer31:

PR #337, opened 2026-09-08
fix(integrations): show why a channel failed, and drop the dead-tab pointer

plombeer31:
> **Stacked on #333 → #332 → … → #327.**

## What was on screen

```
Telegram  · channel failed to start
Discord   · gateway failed — see the Discord tab
```

Three faults in two lines:

1. **"see the Discord tab" points at a tab that does not exist** — removed in #332 when Discord moved into the hub. I fixed this same string in the `configured` branch and missed the `down` branch.
2. **Both lines throw away the reason.** Each channel knows exactly why it failed and exposes it via `lastError()`; the hub simply wasn't reading it. *"channel failed to start"* tells the operator nothing the red badge didn't already say.
3. **The most common cause read like a crash.** A second atomic-agent holding the lock surfaced as `telegram lockfile held by live pid 123` rather than "you already have one running".

## Fix

`IntegrationStatusContext` gains `channelErrors`, populated from `lastError()`, and both descriptors render it:

```
Telegram  · another atomic-agent (pid 42) is already running the Telegram channel — stop it first
```

Lock messages now name the cause **and** the fix.

## Also: inbound Discord logging

An `info` log on every inbound message, emitted **before** any drop decision. When a bot "doesn't answer", the first question is whether the message reached the gateway at all or was deliberately ignored (not addressed, not the owner) — previously indistinguishable from outside.

That log earned its place immediately: it's what proved the Discord gateway was healthy and the real fault was a **stale lock left by a dead process**, not the channel.

```
discord: message received {channelId:"…", authorId:"…", isDm:false, mentionsBot:false}
```

## Testing

`npm run lint` and `npm test` green (7320 tests). New cases pin that the real reason is surfaced, that a plain fallback is used when the channel has none, and — across every channel state — that **no status line ever mentions a Discord tab again**.

### DIFF
--- a/src/channels/discord/discord-channel.ts
+++ b/src/channels/discord/discord-channel.ts
@@ -160,6 +160,17 @@ export class DiscordChannel {
       return;
     }
     if (type !== "MESSAGE_CREATE" || this.botUserId === null) return;
+    // Logged before any drop decision: when a bot "does not answer",
+    // the first thing to establish is whether the message reached us at
+    // all (gateway/intents) or was deliberately ignored (not addressed,
+    // not the owner). Without this the two are indistinguishable.
+    const msg = data as DiscordMessageEvent;
+    this.deps.logger.info("discord: message received", {
+      channelId: msg.channel_id,
+      authorId: msg.author?.id,
+      isDm: msg.guild_id === undefined,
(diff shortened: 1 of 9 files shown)


plombeer31:

PR #338, opened 2026-09-08
fix(integrations): a channel running in another process is not an error

plombeer31:
> **Stacked on #337 → #333 → #332 → … → #327.**

## What was on screen

Running the bots in one terminal and opening the TUI in another — a normal workflow — painted the tab red:

```
Telegram  · another atomic-agent (pid 77052) is already running the Telegram channel — stop it first
Discord   · another atomic-agent (pid 77052) is already running the Discord channel — stop it first
```

Nothing was broken. **Both bots were working**; they were just being served from the other process. Red that fires on a healthy system is red people learn to ignore — which costs us the times it does mean something.

This is a follow-up to #337: that PR made the failure reason *visible*, which then made it obvious the reason wasn't a failure at all.

## Fix

The channel in the second process genuinely cannot start, so `down` remains the right *channel* state. The misjudgement was the **hub treating every `down` as a failure**.

Lockfiles now tag the reason with a stable `channel-locked:` prefix ([`src/channels/channel-lock-error.ts`](src/channels/channel-lock-error.ts)) so the hub can tell a lock conflict from a real failure without pattern-matching on prose, and renders it as an ordinary configured row:

```
Integrations  3/3 configured

  Composio  · connected
  Telegram  · already running in another atomic-agent (pid 6652)
  Discord   · already running in another atomic-agent (pid 6652)
```

A rejected token or disallowed intents **still badge red** — that's where red is the right answer, and a test pins the difference.

## Testing

`npm run lint` and `npm test` green (7327 tests). New cases: the marker round-trips, a real failure is never mistaken for a lock conflict, the machine prefix never reaches the screen, and each descriptor downgrades a lock conflict while keeping a genuine failure at `error`.

Verified against the real thing — bots holding both locks, TUI opened on the same profile, screenshot above.

Also repairs a stray `});` in the two descriptor test files that had been closing `describe` early, leaving later cases running as top-level tests.

### DIFF
--- a/src/channels/channel-lock-error.test.ts
+++ b/src/channels/channel-lock-error.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  CHANNEL_LOCKED_PREFIX,
+  describeChannelLockConflict,
+  formatChannelLockHeld,
+  isChannelLockConflict,
+} from "./channel-lock-error.js";
+
+describe("channel lock conflict", () => {
+  it("tags the reason so it can be told from a real failure", () => {
+    const reason = formatChannelLockHeld(4242);
+    expect(reason.startsWith(CHANNEL_LOCKED_PREFIX)).toBe(true);
+    expect(isChannelLockConflict(reason)).toBe(true);
+    expect(reason).toContain("4242");
+  });
+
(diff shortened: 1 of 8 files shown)


plombeer31:

PR #342, opened 2026-09-08
fix(integrations): make the hub's saves reach the running channels

plombeer31:
> **Stacked on #338 → #337 → #333 → … → #327.** Retarget down the stack as each lands.

## What happened

Found while smoke-testing the RC build, doing exactly what the hub tells you to do:

1. Paste the bot token into Integrations → Telegram. Row says `token saved — press p to pair`.
2. Press `p`. The pane says **"Pairing — DM your bot now; the next sender becomes the owner."**
3. DM the bot. Nothing. No owner, no messages, no reaction — and no indication why.

There was no poller running at any point. `<stateDir>/telegram.lock` never appeared.

## Four faults, one family: a saved value that never reaches the running channel

- **The token never gets to the channel.** `TelegramChannel` resolves its token once, in the constructor (`resolveTokenFromDeps`). The hub writes `.env` itself and then calls `restart()` — and `restart()` only re-starts a channel that was already `up`, which a channel booted without a token never is. So on a fresh install the paste changed nothing in the running process; the channel kept a `null` token and `start()` kept landing in `down` with `missing TELEGRAM_BOT_TOKEN`. The old Telegram tab did not have this bug: it went through `channel.setToken()`, which updates the live value. #329/#332 moved the write to the generic field path and the live half was lost.
- **`p` announces a window that never opened.** `dispatchAction` returns the optimistic string unconditionally while `startPairing()` has already bailed on `channel.state() !== "up"` — its only complaint is an info line the operator is not looking at. Worst case of the four: it tells you to go do the one thing that cannot work.
- **Discord's kill switch was a boot-time snapshot.** `start()` gates on `this.deps.enabled`, so the hub's Channel toggle wrote `discord.enabled: true` and the live channel stayed `disabled` until the next launch.
- **The config cache went stale.** `writeTelegramSettings` never called `resetConfigCache()`, so after `setEnabled` the hub could render `Channel: off` next to a channel that was up — and the next toggle acted on that stale value.

Everything worked after a relaunch, which is what made it look like the hub "sometimes" worked.

## Fix

Each field goes to the mutator that owns the state; `restart()` is not one of them.

| | |
|---|---|
| `TelegramChannel.adoptTokenFromEnv()` | the seam for a writer that has already persisted the token and only needs the running channel to catch up. `setToken` stays the path for writers that want the persistence too |
| `TuiTelegramOrchestrator.adoptToken()` | adopt + start; a cleared token stops the channel. Deliberately **not** the connect chain — that ends in a 60 s pairing window, and the hub offers pairing as its own action |
| `TuiTelegramOrchestrator.ensureUpForPairing()` | `p` starts the channel and, when it can't, throws the channel's own `lastError()` — so the pane reads `401: Unauthorized` instead of "DM your bot now" |
| `DiscordChannel.setEnabled()` | a live kill switch, guarded so a second gateway is never built on one token (two gateways on one token run every turn twice) |
| `writeTelegramSettings` | resets the config cache |

Saving a token now enables the channel, matching what the removed tab's connect chain did — but it stops there rather than opening a pairing window nobody asked for.

## Testing

`npm run lint` clean; full suite **7345 passed**, 1 failed (`sidecar/send-message-concurrency`, fails identically on clean `main` — same baseline #334/#335/#336 report).

`src/tui/integrations/integrations-orchestrator.test.ts` is **new**: the hub orchestrator had no test of its own, which is how this shipped. It drives a **real** `TelegramChannel` behind a fake bot factory, so "did the channel actually come up" is a real answer rather than a mocked one:

- a token saved in the hub starts the running channel
- clearing the token stops it
- `pair` brings the channel up first
- `pair` on a channel that can't start reports the channel's reason and does **not** announce a window
- the Channel toggle starts and stops the live channel
- a missing channel surfaces as an error rather than a silent no-op

Plus `DiscordChannel` (3 cases: `setEnabled` on/off and no second gateway) and `TelegramChannel` (adopting a token written after construction, with the `missing TELEGRAM_BOT_TOKEN` control).

**Non-vacuous**: with the five source files reverted onto the branch, **all 10 new cases fail**.


plombeer31:

PR #343, opened 2026-09-08
feat(integrations): put the setup walkthrough in the hub

plombeer31:
> **Stacked on #342 → #338 → … → #327.** Retarget down the stack as each lands.

## Why

The hub asks for a "Bot token" and leaves you there. Every credential it collects lives on someone else's website — a @BotFather chat, Discord's developer portal, a Composio dashboard — and the person who needs to know that is standing in this pane. A README they don't know exists is not a setup surface.

Watching a first-time setup, the three things that actually blocked progress were all knowable in advance and none of them were on screen: Discord's **client secret is not the bot token**, the **privileged intents must stay off**, and **a bot has to be invited to a server before anyone can DM it**.

## What

Descriptors gain `setupSteps` — one line per step, rendered under the title **until the integration reports `connected`**, so a working setup is never nagged at. The list view points an untouched row at it.

```
Telegram  · no bot token
  https://core.telegram.org/bots#botfather

  Setup
   1. In Telegram, search for @BotFather and open the chat with it.
   2. Send /newbot, then a display name, then a username ending in `bot`.
   3. BotFather replies with a token like 1234567:AA… — copy the whole line.
   4. Press e on Bot token below, paste it, press enter. The channel starts by itself.
   5. Press p, then send your bot any message within 60s — that claims you as the owner.

> Bot token  —
    The 1234567:AA… line @BotFather sends you. Saving it starts the channel.
  Owner  —
  Channel  —
```

Discord gets nine steps (portal → token → intents off → OAuth2 URL Generator → invite → user id → paste → DM). Composio gets four, ending at the sign-in link the first tool call returns. Field help lines are rewritten in the same spirit.

Also: **Discord flips to `appliesLive: true`**. Since #342 stopped the hub relying on `restart()`, its token, kill switch and owner all reach the running channel — the pane was telling operators to restart for no reason.

## Testing

`npm run lint` clean; `src/integrations src/tui/integrations` 91 tests green.

- three panel cases: steps render in detail, the list nudges an untouched row, and both disappear once the row is `connected`
- a registry invariant: every integration with a required secret carries steps, each ≤100 chars so the pane reads as steps rather than a wall

Rendered at 100 columns and read back — no step wraps.


plombeer31:

PR #345, opened 2026-09-08
feat(agent): wait out a provider outage instead of failing the turn

plombeer31:
> **Stacked on #344.** Retarget to `main` once that lands. Implements B1 + B2 from the silent-stop analysis.

## What the trace showed

The provider went unreachable mid-task. The turn died at step 14 after **107 seconds of browser work**, and the next **nine** messages each failed in about a second — across two app restarts and a brand-new session — with nothing on screen saying the link was down. The HTTP client's three fast retries (~1 s) are right for a blip and useless for an outage lasting minutes.

## B2 — park and resume

`agent.providerWait` (`enabled: true`, `maxWaitMs: 300000`) parks the turn instead of failing it: the **same step** is retried on a 2 s → 30 s backoff, and the run continues the moment the provider answers.

- **No replay, no double side effects.** A completion failure throws *before* any tool is dispatched — tool failures come back as results, not throws — so retrying the step re-issues only the completion. `stepsTaken` does not move while parked, so a parked turn cannot eat the task budget from #344.
- **Only failures that plausibly recover are waited on:** no HTTP response at all (DNS, refused, TLS, reset), 5xx, 408, 429. A **404** from a wrong `localModels.url` and a **401** from a dead key are `transport` too — they classify that way so fallover works — and they still fail immediately. Parking a turn for five minutes in front of a typo is worse than the failure it replaces.
- **Esc ends it at once** — the sleep is abort-aware and the turn settles `cancelled`, not `failed`.
- **The backoff is clipped so the last wait ends exactly at `maxWaitMs`** — the configured budget is the budget you get.
- **The budget resets after a recovery**, so a second outage later in a long task gets its own; the task's wall-clock ceiling bounds the total.

## B1 — say it on screen

`provider_waiting` / `provider_recovered` drive a sticky `providerOutage` state:

```
» provider not answering (fetch failed) — retrying in 4s, waited 2s of 300s · Esc stops
```

- **one feed line per outage, not per retry** — the backoff fires every few seconds at first, and a wall of identical lines would bury the work above it;
- the composer meta-row carries the live figure: `waiting for provider 14s/300s — fetch failed`;
- when the budget runs out the row **stays** as `provider unreachable — fetch failed` until a turn actually succeeds. Nine identical one-second failures with a clean screen between them is how one outage reads as eight separate surprises.

## The context readout is untouched

It is driven by `prompt_built` (estimate) and `llm_completed` (the provider's own count). A parked turn builds no new prompt and receives no completion, so the gauge holds its last measured value — it does not reset, move or disappear. There is a test that says exactly that, and the outage indicator lives in the meta-bar's **left** slot (sharing with the composer notice, outranking it) while `contextSlot` and `modeSlot` are separate props at the far end.

## Config

v53 adds `agent.providerWait`; v52 appended to `SUPPORTED_INPUT_VERSIONS`. `enabled: false` — or the per-call `providerWaitEnabled: false` — restores the old fail-fast behaviour exactly.

## Testing

`npm run lint` clean; full suite green apart from the pre-existing `sidecar/send-message-concurrency` failure.

- **loop (6):** parks and resumes with no replay (one tool run, four completions, two waits, one recovery); gives up after the budget with the backoff clipped to it (`2000, 1000` for a 3 s budget); abort during a wait returns in <1.5 s as `cancelled`; a 404 does not park and fails in <1 s; a 503 does park and then succeeds; `providerWaitEnabled: false` fails fast with no events.
- **reducer (6):** shown once with the retry figure; not repeated on retries 2 and 3; cleared on recovery; sticky (`givenUp`) after the turn fails; cleared by a successful `turn_finished`; **context gauge unchanged across park → retry → recover**.
- **readout (4):** budget wording while parked, different wording once given up, ≤80 chars with a long reason, multi-line reason flattened.

Two existing `format-agent-error-for-chat` tests caught the 404 case during development — they drive a real turn against a llama endpoint that 404s and would have hung for the whole wait budget. That is what the waitable/non-waitable split above is for.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -141,6 +141,36 @@ Locked invariants (pinned by [src/agent/agent-loop.test.ts](src/agent/agent-loop
 6. **The closing message names the ceiling, the work done and the way onward** (`formatTaskStoppedReply`),
    and `lastError` carries `task_stopped:<cause>` for post-mortem tooling.
 
+### Waiting out a provider outage
+
+A `transport` failure used to end the turn after the HTTP client's three fast retries (~1s). A field
+trace shows what that costs: the provider went unreachable mid-task, the turn died at step 14 after
+107 seconds of browser work, and the next **nine** messages each failed in about a second — across
+two app restarts and a fresh session — with nothing on screen to say the link was down.
+
+`agent.providerWait` (`enabled` true, `maxWaitMs` 300 000) parks the turn instead. Locked invariants
+(pinned by [src/agent/agent-loop.test.ts](src/agent/agent-loop.test.ts) and
+[src/tui/agent-event-reducer.test.ts](src/tui/agent-event-reducer.test.ts)):
+
+1. **The same step is retried, never a new one.** A completion failure throws before any tool is
+   dispatched — tool failures come back as results, not throws — so replaying the step replays no
+   side effect. `stepsTaken` does not move while parked, so a parked turn cannot eat the task budget.
+2. **Only failures that plausibly recover are waited on.** No HTTP response at all (DNS, refused,
+   TLS, reset), 5xx, 408, 429. A 404 from a wrong `localModels.url` and a 401 from a dead key are
+   `transport` too — they classify that way so fallover works — and they fail **immediately**:
+   parking a turn for five minutes in front of a typo is worse than the failure it replaces.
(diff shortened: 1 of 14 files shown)


plombeer31:

PR #355, opened 2026-09-08
feat(local-llm): configurable Hugging Face endpoint for model downloads

plombeer31:
## What
`localModels.download.hfEndpoint` (config v53) and the `HF_ENDPOINT` env var — the variable `huggingface_hub` itself honours, so anyone already pointing Python tooling at a mirror gets the same behaviour here — send every Hugging Face request to that origin: the repo tree listing behind "add a model from Hugging Face" and the file downloads for catalogue, custom and embedding models. Value is normalised to an origin (`https://hf-mirror.com/` → `https://hf-mirror.com`); anything that is not an http(s) origin is rejected at config load with a message naming the key.

Catalogue entries, custom-model definitions and download sidecars keep their canonical `https://huggingface.co/...` URLs. The origin is swapped at request time only (`rewriteHuggingFaceUrl`), so a partial download survives switching mirrors in either direction, and a GitHub release asset or a URL the operator typed themselves is never rewritten. The `HF_TOKEN` bearer header is sent for canonical HF URLs and for the active endpoint's URLs, as before for huggingface.co alone. The foreground `models pull` line names the mirror host when one is active; `huggingface-api.ts` error messages name the host actually contacted.

## Why
Per-connection throttling (#354) is one reason pulls are slow; the other is regions where huggingface.co itself is slow or unreachable, where a mirror is the only fix. Config for a durable choice, env for a one-off shell.

Stacked on #354 (`feat/parallel-model-downloads`): it edits the same attempt module and continues its config version. Will be retargeted to `main` once #354 merges.

## How it was verified
- `npm run lint` clean
- `npx vitest run src/local-llm src/config src/cli` — 57 files / 752 tests green (baseline failures: none)
- New tests: endpoint normalisation and rejection; env wins over config, unusable env falls back; canonical and `hf.co` URLs rewritten, GitHub untouched, no rewrite without a mirror; the repo listing hits `<mirror>/api/...`; a download goes to the mirror while the sidecar records the canonical URL, and the same partial is then resumed from huggingface.co after the mirror is switched off; config v52 → v53 migration, normalisation, rejection of non-origins


plombeer31:

PR #358, opened 2026-09-08
feat(tui): /report files a GitHub issue with the session's logs, at a privacy level you choose

plombeer31:
## What

Before: when a turn failed, the operator's options were `/dump` (a zip under `~/Documents`) and a hand-written issue.

After: every failed turn ends with `/report files this on GitHub with your logs — you choose how much is shared`. The command (also Help → "Report an issue on GitHub…", `ctrl+g b`) opens a popup with three levels of disclosure, spelled out on the same screen:

| Level | What goes | What never goes |
|---|---|---|
| **Errors only** | version, platform, model, the failures, run outcomes, warn/error log lines, the failure-shaped trace rows | conversation, tool arguments and results, the feed, session ids; every absolute path is masked |
| **Logs, scrubbed** | plus all logs, the feed (tool rows cut to the tool name), run history, the conversation's *shape*, session traces with content fields → `<removed>` | conversation text, reasoning, prompt tails, completions, tool args/results, session ids; home → `~`, cwd → `<cwd>`, emails, IPs, URL queries and credentials redacted |
| **Everything** | the full debug bundle | token-shaped strings — masked at every level |

The zip is written to `~/Documents/atomic-agent-debug/atomic-agent-report-<ts>-<level>.zip` **before** the confirm step, which names the file, its size, the level, the destination (`AtomicBot-ai/atomic-agent`, a public issue under the operator's account) and how many pages the report needs. Nothing reaches GitHub until Enter/`y`. The report is packed into the issue body plus up to three comments within GitHub's 65,536-character limit; the tail of an oversized section is kept, the rest is named as zip-only. Sending needs the GitHub token from the Integrations hub and refuses otherwise, pointing there.

## Why

The operator asked for a crash-report button "with several levels of security". A bug report is a trade between reproducibility and what leaves the machine; that trade is theirs to make per report, with the consequences on screen — not a default in config. The design calls are pinned by tests and written up in AGENTS.md §"Issue reports (`/report`)":

- the zip first, the question second, GitHub last; a leg closed mid-flight drops its result rather than reviving the popup
- redaction is the level's job: `errors` drops free text and masks paths, `scrubbed` scrubs every string (keys included) and strips content fields from traces, `full` masks secrets only; unparsable trace rows are dropped, never passed through
- the popup is modal *above the approval keys*, so `y`/`n`/digits typed at it can never answer a pending approval; Ctrl+C still quits

**Stacked on #357** (`feat/github-integration`): it uses `GithubApi` and `GITHUB_TOKEN` from that PR and cannot compile without it. Base is that branch; I will retarget to `main` when it merges.

## How it was verified

- `npm run lint` clean
- `npx vitest run --no-file-parallelism src/tui/issue-report src/tui/agent-event-reducer.test.ts src/tui/app-key-bindings.test.ts src/tui/backdrop-dismissal.test.ts src/tui/menu src/tui/commands src/tui/submit-handler.test.ts src/tui/tui-app.test.tsx src/tui/mouse src/tui/section.test.ts` — 29 files / 452 tests green (baseline failures: none)
- live (PTY + pyte, 120×36): `/report` opens the picker; ↓ moves the cursor; Enter on "Logs, scrubbed" writes the zip and shows the confirm screen with title, level, destination, page count and zip path; Esc closes; picking "Errors only" and pressing `y` without a token shows "GitHub is not connected. Open the Integrations tab…" with the zip path kept
- an adversarial review pass found tool-result summaries and the user message passing through trace rows, tool payloads in feed lines, session ids in file names and titles, and keys reaching a pending approval; all fixed in the second commit with tests that pin each one


plombeer31:

PR #360, opened 2026-09-08
feat(models): ask once, then ping Telegram or Discord when a background download lands

plombeer31:
## What

People start a 20 GB pull, close the lid, and forget. Nothing tells them it landed, so many never open the app again. This PR makes a background download able to say so.

- **A one-time question when a pull starts.** On the LLM pane, the moment Enter starts a download, a modal asks `Tell you when <model> lands? t Telegram · d Discord · n no · Esc not now`, with the download's progress shown inside it (the modal replaces the pane, so it carries the bar). The answer is remembered in config v52 `notifications.downloads.channel` (`null` = not asked yet); later pulls arm it silently with a `Telegram ping armed for <model> · N to change` line. Esc answers nothing and asks again next time. `N` on the LLM pane reopens the prompt — for the download in flight, or to change the default for the next ones (the remembered pick is ticked).
- **No credentials? Straight to setup.** Choosing Telegram or Discord with no bot token / owner drops the operator into the Integrations hub on that integration's row (`Set up Telegram to get pinged when <model> lands — the download keeps going meanwhile`), via a new `integrations_selected` action. The choice is remembered anyway, and the watch loop arms the ping by itself the moment the hub reports the channel ready — no second trip. A request that never got its credentials is forgotten when its download ends.
- **The worker sends the message itself.** The request rides in `<dataDir>/downloads/<job>.notify` — a sidecar, not a record field, because the worker rewrites the record from memory every few hundred ms and would clobber anything armed after launch; a relaunch onto a partial keeps it. At the end of the job the detached worker (`models pull-worker`) reads it and sends **one plain-text message**: Telegram through a 60-line `fetch`-backed `TelegramApi` shim on top of the existing `sendOutbound` (chunking, 429 retry — no grammy, no `getUpdates`, no lockfile: a request, not a subscription); Discord through `DiscordApi.createDmChannel` + `sendMessage`. `✅ Model ready: Qwen 3.5 4B GGUF — 4.1 GB downloaded — open Atomic Agent to use it` / `⏸ Download paused … 1.0 GB of 4.1 GB is kept — relaunch Atomic Agent and it resumes by itself` / `❌ Download failed: …`. A cancel sends nothing.
- **The ping happens before the terminal record write**, through a new `beforeFinish` hook on `runDownloadWorker`, and its outcome is recorded in the job (`notified: {channel, outcome, reason, at}`). The TUI removes record and log the instant it lands a job, so anything that must outlive the job has to be in the same write that ends it; the TUI then says `Telegram ping sent for …` or `Telegram ping not sent for … — Unauthorized`. Every reason goes through `scrubErrorMessage`; the token is read from the env only to be handed to the sender.
- **CLI:** `models pull --background --notify telegram|discord|off <id>` (also on a pull that is already running — it re-arms the live job), `models downloads` shows `→ telegram` while armed and `→ telegram ✓` / `→ discord ✗ Unauthorized` afterwards. `--notify email` is refused until the Atomic Mail integration lands; `--notify` without `--background` is refused rather than silently ignored.
- Adopted jobs (boot-time, other terminals) are never asked about and never have their CLI-armed ping overwritten. Onboarding's own pull is not asked either.

Stacked on #356 (needs its `waiting`/`resumable` record fields and the worker hooks); will retarget to `main` when that merges.

## Why

The download is the moment we lose people: it is long, it is silent, and it ends when nobody is looking. One question at the start, one message at the end. Telegram and Discord because the hub already holds those credentials — the ping costs the operator nothing they have not already set up, and if they have not, the prompt takes them there.

Invariants kept: the Telegram token stays inside `src/channels/telegram/` + the one-shot path (AGENTS.md §"Telegram remote-control channel" updated); the channel lockfiles guard polling/gateway and are not taken by a one-shot send; the worker never throws for a failed ping (the hook is guarded and logged); `cancelled` gets no message; every file stays ≤300 lines (`download-notify-file.ts`, `download-job-staleness.ts`, `local-models-notify-keys.ts`, `notify-prompt-box.tsx` split out).

## How it was verified

- `npm run lint` clean.
- `npx vitest run src` — 682 / 683 files green; the one failure is `sidecar/send-message-concurrency`, which fails on macOS on clean `main` too and passes in CI.
- New tests: one-shot sender (URL, body, no parse_mode, 429 shape, never throws); notifier (text variants, not_configured paths, Telegram DM, Discord DM channel + message, transport failure → `failed`, readiness); sidecar arm/read/disarm/garbage; spawner arms / keeps / disarms; worker `beforeFinish` runs before the terminal write (asserted: the record on disk is still `running` inside the hook) and a throwing hook cannot fail the download; CLI `--notify` parsing (missing value, `email`, no `--background`), live-job re-arm, listing with outcomes, and the worker actually posting to the Bot API and recording `notified`; reducer + `integrations_selected`; key layers on both the Models tab and the live LLM pane (`n` under the prompt is "no", not Add Provider); orchestrator: asks once, not during onboarding, not after an answer, arms the download in flight, arms remembered channel by itself, hub hop + arming when credentials arrive mid-download, CLI-armed sidecar kept on adoption, pending request forgotten when its download ends, `N` labels; config v52 defaults, v51→v52 migration on disk, bad value rejected.
- Adversarial review pass found the dead `N` (bound only on the unreachable Models tab), adoption re-prompting and overwriting CLI-armed pings, a pending request leaking to the next download, `--notify email/none/no --background` gaps, three files over 300 lines and the missing AGENTS.md notes; all fixed above.
- Live, built `dist/`, real detached worker, throttled local origin:
  - `models pull --background --notify telegram custom-live-model` with a fake token → `models downloads`: `running (pid …) 51% … → telegram`; worker log ends `notify telegram failed: telegram: failed to send chunk (Unauthorized)` **before** `done`; record carries `notified {channel: telegram, outcome: failed, reason: …(Unauthorized)}`; listing shows `→ telegram ✗ … (Unauthorized)`. Same path with a real token delivers.
  - TUI over a real PTY (pyte): Enter on the model → modal `✦ Tell you when custom-live-model lands? / ⇣ 1% · the download keeps going meanwhile / t Telegram · d Discord · n no · Esc not now`; `t` with no token → Integrations hub opens on **Telegram** with `Set up Telegram to get pinged when custom-live-model lands — the download keeps going meanwhile.` while the header chip keeps counting (39%); `t` with a token → `.notify` = `telegram` on the running job and config remembers `telegram`; `N` reopens the prompt with `t Telegram ✓`.


plombeer31:

PR #361, opened 2026-09-08
feat(integrations): Atomic Mail — the agent's own inbox, a verified owner address, and the e-mail leg of download pings

plombeer31:
## What

The third leg of "tell me when it lands": **e-mail, with nothing to set up but your own address.** The agent registers an inbox of its own at Atomic Mail (atomicmail.ai — an e-mail provider built for agents: registration is a proof-of-work, not a signup form), proves that your address is yours with a six-digit code, and from then on can mail you. The inbox is a proper Integrations-hub tenant that later work (an `email.inbox` / `email.send` tool pair, next PR) builds on.

- **`src/atomic-mail/`** — a small hand-rolled client rather than a dependency on the alpha SDK: `atomic-mail-auth` (scrypt proof-of-work `N=16384 r=8 p=1` with the service's fixed salt; challenge → session (1 h) → capability (2 min) JWTs from `Authorization` headers), `atomic-mail-client` (JMAP session discovery, `Mailbox/query`, send = `Email/set` + `EmailSubmission/set` in one batch, inbox = `Email/query` + `Email/get`), `atomic-mail-batches` (the pure wire shapes), `atomic-mail-store` (`ATOMIC_MAIL_API_KEY` in `<stateDir>/.env`; the hour-long session cached at `<stateDir>/atomic-mail/session.json` 0600 so a send inside the hour costs no proof-of-work; the `atomicMail` config block, **config v53**), `atomic-mail-service` (register · sendCode · verifyCode · sendDownloadMail · send · listInbox · forget — the one object every caller uses).
- **Hub row "Atomic Mail"** (`src/integrations/atomic-mail-integration.ts`, logic in `integrations-orchestrator-atomic-mail.ts`): `r` registers `atag-<6 hex>@atomicmail.ai` fire-and-forget (like Telegram pairing, one at a time — the key comes back exactly once) and reports the address; saving *Your e-mail* mails a code there **and only then, atomically with the successful send, records the address as the unverified owner** (the hub applies the field *before* writing it; a failed send changes nothing); *Verification code* is a new **`transient`** field kind — collected by the ordinary edit flow, handed to the orchestrator, stored nowhere — and *Agent inbox* a new **`readonly`** kind. Config keeps `sha256(code)`, a 10-minute expiry and a wrong-guess count (five, then the code is dropped); `ownerVerifiedAt` is set only when the digits come back. `v` resends, `x` forgets the inbox on this machine; pasting or clearing the key drops the cached session and re-discovers the address, so mail never leaves from an inbox whose key was removed. A 401 on a cached session logs in again once instead of failing for the rest of the hour.
- **The mails** (`src/atomic-mail/templates/`): a CRT bezel, a 5×5 block-glyph pixel font in a `<pre>`, two invaders of our own drawn as tables of cells, a readout panel, one line of arcade/RTS-flavoured status in original wording — *Model ready · Additional weights acquired*, *Signal lost · Not enough bandwidth · the partial waits*, *Game over · the file, not the link*, *Access code · Identify, commander*. Tables + inline styles + a Courier stack only, because Gmail and Outlook strip `<style>`, web fonts and scripts; a solid ground under the scanline; a plain-text twin carrying every fact. No user path, no token, everything from the record escaped. Design preview: the four mails rendered from these exact templates — https://claude.ai/code/artifact/08ea0699-ae2b-48cd-8ac5-121caeae6496
- **Notifier**: the `email` channel is ready only with an inbox, an owner address and `ownerVerifiedAt`; the prompt gains `e` and hops to the Atomic Mail row when nothing is set up; `models pull --background --notify email` works.

Stacked on #360 (needs the notify prompt and the worker's `beforeFinish` hook).

## Why

Telegram and Discord each need a bot; e-mail needs an address everyone already has. The one thing that must never happen is mail to someone who did not ask — hence the code, and "nothing is ever mailed to an unverified address except the code itself" (now an AGENTS.md invariant, with a §"Atomic Mail" section and a module-table row). Said honestly there too: the code guards against a typo or someone else's address, not against whoever can edit `config.json` — the gate runs in their own process.

**Known limits, stated in the hub copy where they matter:** atomicmail.ai is an open alpha (free; accounts migrate to the free tier later); outgoing mail gets a "Sent from 🔒 Atomic Mail" footer the service appends; deliverability to Gmail/Outlook is theirs, not ours — the code mail is the first test of it for every operator, and `v` resends.

## How it was verified

- `npm run lint` clean; `npx vitest run src` — 688 / 689 files green (the one failure is the known macOS-only `sidecar/send-message-concurrency`).
- An adversarial review pass found the owner address being written before the code mail went out (a failed send would have left an unverified address "verified"), no recovery from a refused cached session, a second `r` orphaning the first inbox's key, a replaced key still sending from the old session, a status line that said "code sent" when none was pending, no wrong-guess limit, Outlook collapsing the empty sprite cells, `EmailSubmission/set` without `identityId`, and the orchestrator file crossing 300 lines — all fixed above (`integrations-orchestrator.ts` is 356 lines: it was 316 before this PR; the Atomic Mail block lives in its own file).
- New tests: proof-of-work finds a nonce with the required zero bits; client register / login / send batch (exact `Email/set` + `EmailSubmission/set` shape, capability cached) / inbox parse / error hints / session freshness; service register (key → `.env` + process env, address → config, session cached), code mailed + only its hash stored (asserted: the digits appear nowhere in config), wrong / right / expired code, five wrong guesses drop the code, a failed code mail leaves the previous verified owner untouched, a refused cached session logs in again once, reconnect after a pasted or cleared key, download mail refused until verified, forget; store; descriptor status walk, action availability, transient/readonly fields; hub orchestrator `r` in the background, e-mail save → code, code check with the error surfaced, verified row; config v53 defaults / v52 upgrade / whole-or-nothing pending block; notifier e-mail leg + readiness; prompt `e`; templates (facts, escaping, inbox-safety, projector wording, six digit cells); pixel font.
- **Live, against the real service** (a throwaway inbox `atag-f9b244@atomicmail.ai`): registration with the proof-of-work took **2 s**; a send on the cached session **1.4 s** with no proof-of-work; `listInbox` read it back alongside Atomic Mail's own welcome mail. Then the whole owner flow with the agent's inbox standing in for the owner: code mailed → wrong code refused → `482 913` accepted → readiness `ready` → the *Model ready* mail sent and listed.
- TUI over a real PTY: `e` on the prompt with no inbox → the Integrations hub opens on **Atomic Mail** with `Set up E-mail to get pinged when custom-live-model lands — the download keeps going meanwhile.`, the three setup steps and `r register inbox`, while the header chip keeps counting.


plombeer31:

PR #371, opened 2026-09-09
feat(tui): Fusion as the fourth "Where it runs" row, orange chip and chat tint

plombeer31:
## What
Adds **Fusion** to the composer's "Where it runs" switch (`ctrl+r`, or a click on the backend word) as its last row: cloud, local and custom are routes; fusion is a mode built on two of them. While fusion is the route the backend word is an **orange chip**, and the composer's frame and ground and the chat bubbles' borders take the same hue. `/runmode local|cloud|fusion`, `/runmode status` and the `ctrl+g 1/2/3` chords pick a mode from the keyboard. `/run` stays an alias of `/chat`.

Picking fusion runs a pre-flight (a keyed cloud provider and a downloaded local model — the same line the row's detail column shows) and then `RunModeOrchestrator.setMode`, the only TUI writer of `llm.runMode`: it persists the mode, the orchestrator pin and `llm.activeTextProvider` in **one write**, hot-applies the provider, re-mirrors the panel and starts the managed worker daemon (nothing else would — auto-start keys on `local-llama` being the active provider, which under fusion it is not). Cloud/local picks out of fusion go through the same write so the two keys move together.

Stacked on #365 (config + resolver). The runtime does nothing with the mode yet: after this PR fusion = cloud orchestrator active, worker leg persisted, daemon up. The `fusion.delegate` orchestrator follows in later PRs.

## Why
The mode needed a home the operator already knows — the switch that already answers "where does this run?" — and a colour that says which mode the chat is in without a label. Orange is derived from each palette's existing `warnStrong` rather than a seventh token: it is already AA-gated as page ink, and the chip and the mixed composer ground are now gated by `theme-contrast.test.ts`. The word is never painted as orange *text* on the rail (the trap `composer-meta-controls.tsx` documents); the chip form is how it reaches the rail.

Over-budget files (`composer-switch-rows.ts`, `composer-switch-popup.tsx`, `composer-meta-controls.tsx`, `chat-log.tsx`) were split before adding to them, per AGENTS.md layout rules. Design notes in AGENTS.md §"Run modes" › "TUI: the fourth Where it runs row".

## How it was verified
- `npm run lint` clean
- `npx vitest run src/tui` — 259 files / 2835 tests green (whole TUI suite; baseline failures: none)
- New: `persist-run-mode.test.ts`, `run-mode-orchestrator.test.ts`, `fusion-preflight.test.ts`, `dispatch-run-mode.test.ts`, `fusion-bubble-tint.test.tsx`; extended rows/popup/meta-controls/activate/key-bindings, `menu-registry.test.ts` (palette + `command` nodes), `slash-command-handler.test.ts`, `theme-contrast.test.ts` (chip + composer ground on all six palettes), `llm-panel-selectors.test.ts`
- live (PTY + pyte, 120×36, truecolor, scratch state dir with a keyed openrouter entry and a local model on disk): `ctrl+r` lists `cloud · local · custom · fusion` with the blocker in fusion's detail column when a leg is missing; Enter on fusion → meta bar reads `[ fusion ] · openrouter · anthropic/cla… ⇄ qwen-3.5-4b` with the chip cells on `#f98f3a` / ink `#13161d`, composer border `#f98f3a`, composer ground mixed to `#674530`; `/runmode status` prints the resolver line; `/runmode local` clears the mode and moves `activeTextProvider` to `local-llama` in the file.


plombeer31:

PR #372, opened 2026-09-09
feat(git): remote sync tools — remote, fetch, pull, push, clone behind the closed-repo switch

plombeer31:
## What

Before: the agent could only read a repository (`os.git.status/log/diff/show/blame/branch`); anything that touched a remote went through `os.shell.run`, where a `git push` was one approval away and a GitHub token could only reach git by being typed into a URL.

After: five dedicated tools — `os.git.remote` (list / add / set-url / remove), `os.git.fetch`, `os.git.pull`, `os.git.push`, `os.git.clone` — built so a closed repository stays closed and a token never leaks:

- **The switch comes first.** Every network verb checks `git.remoteSync` (the Remote sync switch from #368) before the approval gate and before git is spawned. Off ⇒ a structured error naming the switch and stating nothing was sent or fetched. `remote list` is read-only and always works; `remote remove` works with the switch off (it only makes a repository more local). With no policy injected at all, the tools stay closed.
- **One approval category, `git_remote`** (new, label "git · remote"), silent from level 4 — the same rung as a guarded `git push` through the shell, so the dedicated tools are never looser or stricter than the escape hatch. Session-grantable. The prompt preview is the exact git command line.
- **The token never touches argv, a URL or `.git/config`.** `GITHUB_TOKEN` is copied into a variable on the child git process only, and git is pointed at an inline credential helper bound to `credential.https://github.com.helper` — any other host never sees it (proved in the tests with `git credential fill` against real git). With no token git falls back to the operator's own SSH keys / keychain / `gh` helper. `remote add|set-url` and `clone` refuse a URL carrying `user:token@`, for any host.
- **Output is scrubbed** of the token before it reaches the transcript or the trace.
- **Nothing hangs, nothing is forced.** SSH runs in batch mode unless `GIT_SSH_COMMAND` is set; `push` has no force flag and sets the upstream on a branch's first push; `pull` is `--ff-only` unless `rebase: true`; `clone` refuses an existing destination or a missing parent.

**Stacked on #368** (needs `config.git.remoteSync` and the injected `shellPolicy` predicate); retarget to `main` once that merges. The sibling PR with the local write tools (`init`/`add`/`commit`/`checkout`) is independent; both touch `tools/os/index.ts`, `tool-resource-class.ts`, the descriptor and schema tables — expect a trivial adjacent-line conflict when the second lands.

## Why

An operator who keeps a project private should be able to hand it to GitHub only on purpose: flip the switch, approve the push, and know that the token used for it was never written anywhere a `git remote -v` or a shell history would show it. Every design call here follows from that — the switch before the ladder, one category at the shell's rung, host-scoped credential injection, and scrubbed output because the trace records tool results verbatim.

## How it was verified

- `npm run lint` clean
- `npx vitest run src/tools/os/git src/approval src/agent/tool-resource-class.test.ts src/prompt/default-tool-args-schemas.test.ts src/prompt/build-prompt.test.ts src/tui/approval-modal.test.tsx src/channels/telegram/approval-bridge.test.ts --no-file-parallelism` — 18 files / 196 tests green (run serially: this machine was at load 16–28 from other test runs, which times out git-spawning hooks under the default 10 s; baseline failures: none)
- new suites: `git-remote-sync.test.ts` drives a local bare repository through add → push → fetch → pull → clone with real git and records every approval (`git_remote`, exact previews), plus refusal-before-git with the switch off, denial leaves the remote empty, credential-bearing URLs refused, detached HEAD, git's own failures; `git-credentials.test.ts` runs `git credential fill` to prove the helper answers `github.com` and not `example.com`; `git-remote-policy.test.ts` pins the refusal shape and the category.
- live: a CLI transcript over a local bare remote — with the switch off `remote add` and `push` return the refusal without prompting; after flipping it, `remote add` → `push --set-upstream origin main` → `fetch` → `pull --ff-only` → `clone` each raise one `git_remote` approval with the git command line as the preview, the bare repo's `main` equals the local `main`, and a `user:token@` URL is refused without echoing the token.


plombeer31:

PR #374, opened 2026-09-09
feat(tui): Fusion configurators — orchestrator provider, worker model, worker count

plombeer31:
## What
Adds the Fusion configurators to the composer. On the fusion route the strip walks **four** controls — `backend ⇄ provider ⇄ model ⇄ workers` — instead of three.

- **Provider / model** address the **orchestrator** leg. Picking a provider re-pins `fusion.orchestratorProvider` through the run-mode write instead of calling `setActiveText`; picking one of its models keeps the existing path (same provider id, so the pin holds).
- **Workers** (new, fusion-only) is the **local** half: the downloaded models the workers run, then `1..8 workers`. A model pick goes through `LocalModelsOrchestrator.setActive` — it restarts the managed daemon and writes only `localModels.*`, never `activeTextProvider`. A count pick writes `llm.runMode.fusion.workers` and `localModels.managed.parallel` in one write and says how to apply the new slot count.
- `/runmode workers N` is the same call from the keyboard.

Stacked on #371 (composer switch), which is stacked on #365 (config + resolver).

## Why
Fusion is the one route with two legs, so it is the one route that needs a control for each. Keeping the worker count and `--parallel` in a single write matters because they are one fact seen from two sides: how many workers may fan out, and how many llama-server slots exist for them to run in. Split writes let them disagree, and N workers then queue on the old slot count with nothing on screen to explain the lost concurrency — hence the restart notice.

Two traps the tests pin. `setActiveText` under fusion would move the active provider off the orchestrator pin and silently drop the mode on the next read, so the provider control routes through the run-mode write. And the LLM pane's row `active` additionally requires `local-llama` to be the chat route — under fusion it never is — so the worker rows read the daemon's own model flag instead. (The second one was caught by a failing test, not by review.)

## How it was verified
- `npm run lint` clean
- `npx vitest run src/tui` — 261 files / 2864 tests green (whole TUI suite; baseline failures: none)
- New: `composer-switch-worker-rows.test.ts`; extended `composer-switch-activate` (worker model / count / orchestrator re-pin / keyless provider still opens the wizard), `composer-switch-key-bindings` (four-control walk on fusion, `workers` unreachable off it), `composer-meta-controls` (fourth control), `persist-run-mode` (both counts in one write, bounds), `run-mode-orchestrator` (restart notice, no notice when the count did not move), `dispatch-run-mode` (`workers N`, bounds)
- live (PTY + pyte, 120×36, truecolor, fusion stored, one model on disk): meta bar reads `[ fusion ] · openrouter · anthropic/cla… ⇄ qwen-3.5-4b · 2 workers`; `→→→` opens `WORKERS (1/10)` listing `✓ qwen-3.5-4b  worker model` then `1..8 workers` with `✓ 2 workers` and the download deep link; picking `6 workers` updates the bar and writes `runMode.fusion.workers: 6` **and** `localModels.managed.parallel: 6`.
- full suite on this branch: `npx vitest run` — 708/709 files, 7705 tests green; the single failure is `sidecar/send-message-concurrency`, the known macOS-only baseline failure that passes in CI on Linux


plombeer31:

PR #375, opened 2026-09-09
feat(runtime): per-turn provider pin, ephemeral worker sessions, worker approval policy

plombeer31:
## What
Three runtime seams that let one process run a cloud orchestrator turn and several local worker turns at the same time. Nothing user-visible yet — the `fusion.delegate` tool that consumes them is the next PR.

**A per-turn provider pin.** `runTurn(session, msg, { providerId })` pins every completion of that turn to one configured provider: the step is built for *that* link's transport, adapter and slot affinity (the four loop deps otherwise describe the active provider), and the request bypasses the fallback chain entirely. A pinned turn never falls over — the pin exists to spend local tokens, and a fallover would silently run the workers on the cloud leg and invert the cost model the mode was chosen for; the orchestrator is the retry authority, and the chain's breaker state stays untouched by a link it did not pick. An id the registry does not know is rejected before the turn is queued, so a bad pin fails loudly instead of quietly running on the active provider. En route, usage is now priced against the provider that actually **served** rather than the active one, which also fixes pricing after an ordinary cross-provider fallover.

**Ephemeral worker sessions.** `runtime.createEphemeralSession(meta)` mints an in-memory session stamped as a fusion worker: never saved, no trace recorder, no memory recall, no reflection, no lesson bump, and a `"fusion"` turn origin that is treated like `scheduler` wherever origin gates analytics and the usage meter. The orchestrator reads the transcript once and discards it, so nothing reaches the session list or the operator's long-term memory.

**A worker tool policy.** `RunTurnOptions.toolFilter` hides tools from a turn (a worker may not delegate further, schedule durable work, or write memory), and `ApprovalGate.setSessionPolicy(id, { onPrompt: "refuse", reason })` denies — without emitting — any request from that session that auto-approval does not already settle. A worker has nobody at the other end of a prompt (the operator is looking at the parent session), so parking its turn on an unanswerable question would hold it until the process exits; instead the tool result tells the worker to report back what needs doing. Auto/bypass levels still run the call as before.

Plus `SlotManager.poolSize()` — the slots `acquire()` can actually hand out, which is what the fan-out will size its concurrency from.

Stacked on #365 (config + resolver).

## Why
Every one of these is a gap the fusion design hits immediately: the loop had no way to run a step on a provider other than the active one, every `runTurn` persisted its session and appeared in the rail, and an approval prompt from a hidden session would deadlock the fan-out. Each is useful on its own terms too — the pricing fix and the loud rejection of an unknown provider id are not fusion-specific.

Design notes in AGENTS.md §"Concurrency contract".

## How it was verified
- `npm run lint` clean
- `npx vitest run src/agent src/runtime src/approval src/session src/tools/fusion src/llm/slot-manager.test.ts src/llm/provider src/tasks` — 92 files / 1078 tests green (includes `turn-controller`, `task-runner`, `bootstrap`, `agent-loop`); no reruns needed
- New: `agent-loop-fusion-seams.test.ts`, `bootstrap-fusion-seams.test.ts`, `approval-gate-session-policy.test.ts`, `fusion-worker-session.test.ts`, `worker-tool-policy.test.ts`, `resolve-model-pricing.test.ts`, seam cases in `llm-fallback-seam.test.ts`, `slot-manager.test.ts`
- Adversarial pass: every new test was confirmed to fail without its change (pin → the chain picks the cloud model; ephemeral → recalls > 0; `toolFilter` → the tool is still on the wire; unknown providerId → the turn runs; policy → a prompt is emitted)

One nuance found while building it and documented in the code: the OpenAI adapter appends `reply`/`finish` to the wire payload unconditionally, so filtering a terminal verb hides it from the prompt catalog only. Harmless for a worker — calling `finish` ends only its own throwaway session.


plombeer31:

PR #376, opened 2026-09-09
feat(tui): pin sessions — a white ↑ on every row keeps a thread at the top

plombeer31:
## What

Stacked on #369 (`feat/session-rail-reorder`) — please merge that first; this branch is based on it and will be retargeted to `main` once it lands.

Threads can now be pinned to the top of the rail:

- **An up arrow at the right edge of every session row.** Bright and bold when the thread is pinned, dim when it is not. Clicking it pins the row; clicking again releases it. Unlike the `[x]`, the mark is on every row rather than only the selected one — pinning is not destructive, and a mark that only appears under the pointer cannot answer the other half of the question the control exists for: which threads are already pinned. Its two columns are reserved on every row, like the close mark's four, so nothing materialises under a pointer mid-gesture.
- **`p` on the focused row does the same**, for the same reason `x` exists beside the close mark: the mark is painted whether or not mouse reporting is on. The cursor follows the row to its new slot, so pressing `p` twice undoes the first press instead of pinning a second thread.
- **The pinned block sits at the top of the rail**, manually ordered like the rest: pinning appends to the block, releasing drops the row at the head of the unpinned half, and Shift+↑/↓ moves stop at the block's edge — a keyboard move never pins or unpins. A **drag** decides by where it lands: dropped inside the block it pins, dropped below it releases, which is what a desktop list does.
- **A pin outlives the recency window.** When a pinned thread has fallen out of the list `railSessions()` reads, the rail fetches it by id and builds the same row; a pin on a deleted session is pruned on the next write.
- **Remembered** in `tui.sessionRail.pinned` (config v53), beside the v52 `order`. Pinning writes config only — the session row is untouched, so a pin never bumps `updatedAt`.
- The modal picker lists the same arranged list, so pinned rows carry the mark there too.

## Why

The rail is the operator's working set, and a handful of threads are the ones they come back to daily. Recency buries those under whatever ran last; a pin is how every desktop list has solved that.

## How it was verified

- `npm run lint` clean
- `npx vitest run src/tui/session-rail src/tui/reduce-sidebar-actions.test.ts src/tui/reduce-session-actions.test.ts src/tui/app-key-bindings.test.ts src/tui/session-delete-keys.test.ts src/tui/rail-session-list.test.ts src/tui/components/sidebar.test.tsx src/tui/components/sidebar-fit.test.tsx src/tui/chat-orchestrator.test.ts src/config src/tui/mouse/mouse-app.test.tsx` — 26 files / 484 tests green. New: the pin arithmetic (append, release, block edge, drag across the boundary, pruning), `p` (including the cursor follow, upper-case `P`, Tasks pane untouched, other letters still swallowed), a `mouse-app` click on a row's arrow that pins without opening or moving the row, sidebar renders for the mark on every row, config v52→v53 inheritance.
- Live, in a PTY at 120×36 with six seeded threads: `p` pinned the selected row, which moved to the top with the cursor still on it, and `p` again released it — `tui.sessionRail.pinned` in `config.json` followed both. Frame colours checked against the palette: a pinned unselected row's arrow renders `#eef1f6` (`railForeground`) bold, an unpinned row's `#b9c8ea` (`railMuted`).

Two width assertions written against the wider preview moved with the two reserved columns (`sidebar.test.tsx`, and the delete tests in `mouse-app.test.tsx` now match a preview prefix, as the neighbouring row tests already did).

Note: this bumps `USER_CONFIG_VERSION` to 53, which #355/#361 also claim — the RC renumbers whichever lands second.

### DIFF
--- a/src/config/config-schema.test.ts
+++ b/src/config/config-schema.test.ts
@@ -1543,7 +1543,7 @@ describe("tui.sessionRail (config v52)", () => {
   it("gives a v51 file the recency default — an empty order", () => {
     const parsed = parseUserConfigFile({ version: 51, tui: { theme: "nord" } });
     expect(parsed.version).toBe(USER_CONFIG_VERSION);
-    expect(parsed.tui.sessionRail).toEqual({ order: [] });
+    expect(parsed.tui.sessionRail).toEqual({ order: [], pinned: [] });
     expect(parsed.tui.theme).toBe("nord");
   });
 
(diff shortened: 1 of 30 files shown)


plombeer31:

PR #378, opened 2026-09-09
feat(agent): fusion.delegate — the cloud orchestrator fans work out to local workers

plombeer31:
## What
The orchestrator itself. In fusion mode the cloud model plans, then calls **`fusion.delegate`** with a list of self-contained tasks; each becomes an ephemeral worker session running a turn pinned to the llama-server leg, several at a time, and every worker's reply comes back in one tool result for the cloud model to review and merge. Cloud tokens pay for the thinking, local tokens for the bulk.

- **The tool** (`src/tools/fusion/`) takes `{ tasks: [{ id, title, instructions, deliverable?, files? }], maxWorkers? }`. Concurrency is capped by the configured worker count, the task count, and — when the local leg keeps KV-cache affinity — the slots the server actually has; with a pool of one it says so and names `localModels.managed.parallel`. It refuses before doing any work when called from inside a worker (one level of fan-out), when fusion is no longer the effective mode (re-read live, so an operator who switches provider mid-turn is not billed for a leg they walked away from), or on invalid args. Once it runs it returns `ok` even if every worker failed: per-task status lives in the output, so the orchestrator always sees partial results instead of a bare error. It is `approval_gated` in the resource-class table, i.e. always solo in a batch.
- **Workers** get a brief telling them they have no memory of the parent conversation and must report rather than ask. They cannot delegate further, schedule durable work or write memory, and an approval that would prompt is refused with a message telling them to hand the step back — classified as `needs_orchestrator` in the result.
- **Guidance** for the cloud model is a `### fusion` block in the stable prefix, present only when the descriptor is — zero bytes for every non-fusion install.
- **Progress** shows up in the chat as `» worker <title>: started / done — N steps` via a new `fusion_worker` event, emitted in the **parent's** frame so it reaches the operator's session rather than the invisible worker's.

Stacked on #375 (runtime seams) → #365 (config).

## Why
This is the point of the mode: spend the expensive model only on planning, instruction-writing and review, and the cheap local one on the bulk that can be done in parallel. Everything else in the stack exists to make this call safe — a pinned turn that cannot fall over to the cloud leg, sessions that leave no trace, and workers that cannot prompt a human who is looking at a different session.

Two decisions worth flagging. There is **no plan-approval gate**: workers are the orchestrator's subtasks and it delegates whenever that is more efficient. And a worker's `started` event is emitted from inside its turn hook rather than before the call — the adversarial pass found the original placement ran in the parent's async context anyway, which made the parent-frame contract accidental rather than real; moved, the test genuinely fails if the event is emitted from the worker's own context.

## How it was verified
- `npm run lint` clean
- `npx vitest run src/tools/fusion src/prompt src/agent src/runtime src/tui/agent-event-reducer.test.ts src/sidecar` — 52/53 files, 718 tests green. The one failure is `sidecar/send-message-concurrency`, confirmed failing identically on the base branch (known macOS-only; passes in CI on Linux)
- Wider sweep `src/tui src/http src/llm src/tasks src/channels src/tools` — 452 files / 4863 tests green
- Integration (`fusion-delegate.integration.test.ts`, a real runtime with a cloud and a local provider entry): three workers fan out with peak in-flight ≥ 2 through the real `TurnController`; six `fusion_worker` events, every one tagged with the **parent** session id; no `s-w-` row reaches the session store; a worker's `os.fs.write` at approval level 1 is denied with **zero** approval prompts emitted and lands as `needs_orchestrator`; the orchestrator's prompt carries the delegate tool and the `### fusion` block while the worker's carries neither

One gap in #375 surfaced while building this and is fixed here: `toolFilter` existed on the agent loop but was never threaded through `runtime.runTurn`, so nothing outside the loop could actually use it.


plombeer31:

PR #382, opened 2026-09-09
feat(llm): output ceilings you choose — none on cloud by default, none on local if you say so

plombeer31:
## What
Makes the output ceiling a choice instead of an accident, on both legs.

**Cloud keeps no ceiling of its own.** Building on #381, `max_tokens` is sent only when something asks for a bound: the call's own value, then the provider entry's new **`maxOutputTokens`**, else nothing — the service applies the model's own maximum. That is what makes a single long answer possible. Set the entry field to bound spend, or for a service that requires the field:

```json
"llm": { "providers": [{ "id": "openrouter", "kind": "openrouter", "maxOutputTokens": 64000 }] }
```

**Local gains the ability to have none.** `localModels.completionMaxTokens: 0` becomes llama.cpp's `n_predict: -1` — generate until a stop token or until the context window fills, which is the real ceiling on a local run. The default stays 8192 and every other value keeps its 64..131072 window.

Config version 51 → 52.

## Why
The two legs were answering one question with one number, and it was the wrong number for one of them. What a cap means locally and what it means on a cloud service are not the same thing, and the docs now say so rather than leaving it to be inferred.

Worth stating plainly, because it is easy to assume otherwise: **a local cap is not a memory guard.** What the machine commits is decided at daemon start by the model and `--ctx-size`; it does not grow with the length of one reply. A positive `n_predict` bounds wall-clock time and a degenerate loop — real things, worth having a knob for — but nothing about it protects the machine from a long answer, so an operator who wants long local answers can safely turn it off.

And the trap behind the bug this pair fixes: **a reasoning model spends the same budget on thinking.** A ceiling of N is a ceiling on reasoning *plus* output, so a low one can be exhausted before a single byte of content is emitted. A truncated completion with empty `content` and a long `reasoningContent` is exactly that, and now has a documented cause.

## How it was verified
- `npm run lint` clean
- `npx vitest run src/config src/llm src/agent src/runtime` — 110 files / 1533 tests green
- New `resolve-n-predict.test.ts` (a configured or per-call `0` becomes `-1`, everything else passes through); new cases in `config-schema.test.ts` (0 accepted, bounds still enforced, a v51 file keeps 8192), `llm-config.test.ts` (`maxOutputTokens` round-trips, absent by default, rejects non-positive and fractional), and `openai-build-body.test.ts` (the entry ceiling applies, the call's own value wins, neither means no field)

Stacked on #381 — that PR stops the local budget leaking into cloud requests, this one gives both legs a ceiling the operator sets. **Note the config version:** several other open PRs also claim 52; whichever lands second needs a re-bump plus the outgoing number appended to `SUPPORTED_INPUT_VERSIONS`.


plombeer31:

PR #385, opened 2026-09-09
feat(tui): a fallover notice you can act on

plombeer31:
## What
The provider-fallover notice from #384 gains a `[configure fallback]` button beside `[copy]`. Clicking it deep-links to Manage › LLM › Fallback, where the chain is actually edited.

## Why
The notice tells the operator that a provider dropped out and something else is answering — which is a decision handed to them: accept it, put a different provider next, or take the local one out of the chain entirely. The pane that makes that decision is `/llm` plus a pane most operators have never opened. Announcing the change and leaving the fix undiscoverable is half a message.

A deep link rather than a modal over the chat: the chain is edited in the Fallback pane, and a second editor for the same config is how two writers of `llm.fallback` come to disagree. The offer rides on the message as an `action`, the way `retryText` already carries the abort notice's offer, so the footer stays one row and `estimateMessageHeight` does not have to branch on it.

Without a mouse provider the label still renders — the same call `[copy]` makes — so under `--no-mouse` the affordance is legible and `/llm` is the way there.

Stacked on #384.

## How it was verified
- `npm run lint` clean
- `npx vitest run src/tui` — full TUI suite green
- New `chat-configure-fallback-button.test.tsx`: the three deep-link dispatches in order and targeting the Fallback pane, and the label still rendering with no mouse provider
- Extended the fallover case in `agent-event-reducer.test.ts` to assert the notice carries the offer

### DIFF
--- a/src/llm/fallback/describe-reason.test.ts
+++ b/src/llm/fallback/describe-reason.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from "vitest";
+
+import { describeReason } from "./provider-fallback-chain.js";
+
+/**
+ * The reason text is what an operator reads on a fallover notice and in
+ * the feed, so it has to name the refusal rather than the transport that
+ * carried it.
+ */
+describe("the reason an operator is shown for a fallover", () => {
+  it("is the provider's own message, not the error class", () => {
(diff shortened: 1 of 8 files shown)


plombeer31:

PR #450, opened 2026-09-15
desktop: Soft Tactile redesign — every screen, integrated

plombeer31:
The whole Soft Tactile redesign of the desktop in one branch: the foundation and the nine region PRs merged, plus the fixes that only show once they run together. This is the build to try and the one to merge; the region PRs are there to review one area at a time.

## Contents
| PR | Area | Screens |
|---|---|---|
| foundation | tokens, Figtree, brand logos, icons, shared kit, stylesheet split by region | — |
| shell | toolbar, floating sidebar and inspector, console drawer, download strip, rail | SH-01…09, SH-17, CH-23 |
| overlays | palette, sheets, delete alert, toasts, theme that persists | SH-03, SH-10…16 |
| composer | glass composer, Backend · Provider · Model with logos, status line, voice strip, slash list | CM-01…07, CM-09…11 |
| popovers | route / provider / model popovers anchored per chip, wizard, modes, context, voice languages | MP-01…18, CM-08, CM-12…17 |
| transcript | bubbles, tool cards, indigo approval block, plan hand-off, notices | CH-01…26 |
| onboarding | title card, indigo side rail, route cards, logo lists, every wizard step | ON-01…23 |
| settings frame | Manage window and nav, Tasks, Skills, Privacy | ST-01…11, ST-33 |
| settings data | Memory, MCP, Telegram, Import | ST-12…19, ST-28…32 |
| settings LLM | route card, run mode, models, providers, external, fallback, logs | ST-20…27 |

## Fixes that live only here
- **Settings keeps no hidden text, and the smoke reads what is drawn.** To keep the smoke's old terminal-text checks passing, the three Settings PRs kept invisible copies of TUI strings in the DOM (`sdTwin`, `.set-sr`, `.set-z`, `llmSr`). A test that passes on text nobody can see proves nothing, so here they are gone. About thirty smoke checks now assert on the visible screen: nav labels and count badges, table headers, empty-state titles, the channel and mode switches, provider rows with their key chips, Import labels and values. Every state, data and config assertion is unchanged. Proof: renaming three visible labels on purpose failed exactly those three checks, each naming what broke; reverting made them pass again. Text that was only `aria-hidden` behind a copy is readable again, and menu rows carry `aria-keyshortcuts`.
- The Memory and MCP background refreshes no longer overwrite whichever Settings nav row is active.
- `cloud-setup.drive.mjs` reads the provider row (id, key chip, Enter text) instead of bracketed text the redesign no longer draws.
- The Import tab's Source row lines up with the fields below it.
- Kit: an on toolbar toggle sits on `--tint` (follows the theme); `.tk-spin` is inline-block; the destructive fill's label reads in dark. Test infra: the smoke reads every renderer stylesheet; `visual.drive.mjs` checks Figtree for Latin and Inter for Cyrillic; `drive-selector-lib.mjs` launches with a throwaway profile. (These are also on the foundation PR.)

## Verified on this branch
All runs on scratch copies of a configured state (settings version set to the build's, chats moved to a scratch workspace) or fresh directories, each with its own `--user-data-dir` and CDP port — never the operator's state or profile.

| Check | `a0811bc3` | this branch |
|---|---|---|
| token check (`styles.css` + `css/*.css`) | — | clean |
| `electron . --smoke` | 497 pass · 4 fail | 497 pass · the same 4 fail |
| `model-picks.drive.mjs` (68 / 8 / 4 GB) | 23 / 29 | 29 / 29 |
| `drive-selector.mjs` | 3 custom-route fails | the same 3 |
| `unverified.drive.mjs` | — | 7 / 7 |
| `model-step.drive.mjs` (real AI/ML API catalogue) | — | 5 / 5 |
| `failure-line.drive.mjs` | — | 5 / 5 |
| `visual.drive.mjs` | — | Figtree, Inter (Cyrillic) and DM Mono load; slash list painted, 34 rows, unclipped, both themes |
| `onboarding-mouse.mjs`, keyed state, one run | 61 / 61 | 61 / 61 unloaded |
| `onboarding-mouse.mjs`, keyed state, four at once | 2 / 2 fail at the import step (same batch) | 2 / 2 fail at the same step (same batch) |

The four-at-once miss predates this work: the wizard's import step moves on before the driver's click under heavy load, on 0.5.5 as well.

The 4 known smoke failures on both: `llm tab … [downloaded]` (no model on disk in the fixture), `state dir: the import copies providers without keys` (fixture keys inline), `backend: config round-trip to local` (daemon untouched), `switch timings … selectLocalModel=NOT EXERCISED`.

Screens were compared against the AA-DS-002 references at 1470×923 and 940×620 in light and dark.

## Found, not changed
- Choosing Fusion in Settings › LLM reports success but the 0.5.5 agent's config schema has no `llm.runMode`, so the whole-file write drops it; Fusion needs an agent ≥ 0.6.0.
- Cloud reads "Active" on local routes (unset run mode defaults to cloud); "key ok" means a key name is present, not checked.
- Main-process error copy still reaches users as "fetch failed", "export HF_TOKEN", "re-run with --refresh".
- A cancelled task still offers Run now and Cancel; the session sheet's filter and Open do nothing; Enter on the palette's "No results" does nothing (only a click on Ask the agent works).
- `drive-selector.mjs` fails 3 custom-route checks on `a0811bc3` too (no llama-server answering, so no model chip).

## Needs a decision
- The sidebar Settings entry now follows the reference (neutral button, gear, word, ⌘, keycap). That reverses the r5 item 8 wording ("a plain blue button — no keycap hint, no second icon"). Say which one stays.

## Region PRs
- desktop: Soft Tactile foundation — tokens, Figtree, brand logos, icons, shared kit — https://github.com/AtomicBot-ai/atomic-agent/pull/440
- desktop: Soft Tactile window shell — floating sidebar, inspector, console drawer, download strip — https://github.com/AtomicBot-ai/atomic-agent/pull/441
- desktop: Soft Tactile composer — glass card, Backend · Provider · Model with logos, status line, voice strip — https://github.com/AtomicBot-ai/atomic-agent/pull/442
- desktop: Soft Tactile composer popovers — per-chip anchoring, logo rows, wizard, modes, context — https://github.com/AtomicBot-ai/atomic-agent/pull/443
- desktop: Soft Tactile transcript — bubbles, tool cards, indigo approval block, plan hand-off — https://github.com/AtomicBot-ai/atomic-agent/pull/444
- desktop: Soft Tactile overlays — palette, sheets, alert, toasts, persisted theme — https://github.com/AtomicBot-ai/atomic-agent/pull/445
- desktop: Soft Tactile onboarding — indigo side rail and every wizard step — https://github.com/AtomicBot-ai/atomic-agent/pull/446
- desktop: Soft Tactile Settings — window, nav, Tasks, Skills, Privacy as real screens — https://github.com/AtomicBot-ai/atomic-agent/pull/447
- desktop: Soft Tactile Settings — Memory, MCP, Telegram, Import as real screens — https://github.com/AtomicBot-ai/atomic-agent/pull/448
- desktop: Soft Tactile Settings › LLM as real screens — https://github.com/AtomicBot-ai/atomic-agent/pull/449


plombeer31:
The desktop's agent now has its own PR: #451 merges `main` (v0.6.1) into `atagdesktop`. That brings config format 65, which the live desktop state (64) needs, and the `llm.runMode` schema the Fusion settings here rely on. A test DMG built from this branch with that agent passed 7/7 when driven from the DMG copy against a copy of a live state: it connects on the bundled agent, the Tactile UI loads, mode switching goes through `/api/coding-mode`, and a Settings write lands with no version wall.

plombeer31:
Round 2 from testing the DMG is #454, based on this branch. It covers the first-run trims, spinner, model tick and double click, comet, one Stop, no app-status strip, closable toasts, the house icon, and Fusion as a backend mode. The agent's final reply is now the last row of a turn. The agent side of those fixes (fusion_worker SSE, stored approvals, the vision catalogue check, shell results, replies that never stream) is on #451.

